PR #40, opened 2020-12-14
make hexToRgba recognize CSS custom properties

andreiglingeanu:
Solves the problem of applying the opacity onto the colors from the Gutenberg color palette, when the color palette contains a CSS Custom Property. It expects the css variable to be defined on the `:root` element.

How to test:

Apply this palette to Gutenberg:

```php
add_theme_support( 'editor-color-palette', [
    [
        'name' => __( 'Palette Color 1', 'blocksy' ),
        'slug' => 'palette-color-1',
        'color' => 'var(--paletteColor1)',
    ],

    [
        'name' => __( 'Palette Color 2', 'blocksy' ),
        'slug' => 'palette-color-2',
        'color' => 'var(--paletteColor2)',
    ],

    [
        'name' => __( 'Palette Color 3', 'blocksy' ),
        'slug' => 'palette-color-3',
        'color' => 'var(--paletteColor3)',
    ],

    [
        'name' => __( 'Palette Color 4', 'blocksy' ),
        'slug' => 'palette-color-4',
        'color' => 'var(--paletteColor4, #fff)',
    ],

    [
        'name' => __( 'Palette Color 5', 'blocksy' ),
        'slug' => 'palette-color-5',
        'color' => 'var(--paletteColor5, rgba(0, 0, 0, 0.5))',
    ]
]);
```

With that in place, define these CSS variables from:

```css
:root {
  --paletteColor1: red;
  --paletteColor2: green;
  --paletteColor3: blue;
  --paletteColor4: pink;
  --paletteColor5: black;
}
```

With this in place, `hexToRgba()` will be able to apply the opacity correctly. I'll be happy to change the implementation if this somehow doesn't fit the project's guidelines and rules.

---

This approach for the Color Palette is used in [Blocksy](https://wordpress.org/themes/blocksy/) and there were no problems with this up until now. Hope this fix could be included in Kadence Blocks for a better compatibility between the two tools.

A similar PR was applied on [Stackable](https://github.com/gambitph/Stackable/pull/706) and [Kadence](https://github.com/kadencewp/kadence-blocks/pull/106) blocks and runs with success so far.

---

_Note_: Kadence and Stackable lack the support for defaults and rgba values in defaults. The code in this PR offers this. Happy to update the implementation as the review moves ahead.

### DIFF
--- a/src/utils/hex-to-rgba/index.js
+++ b/src/utils/hex-to-rgba/index.js
@@ -34,7 +34,6 @@ export default function hexToRGBA( hex, alpha ) {
 		}
 	}
 
-
 	if ( ! alpha && 0 !== alpha ) {
 		return hex;
 	}
--- a/src/utils/hex-to-rgba/index.js
+++ b/src/utils/hex-to-rgba/index.js
@@ -10,6 +10,31 @@ export default function hexToRGBA( hex, alpha ) {
 		return '';
 	}
 
+	/**
+	 * Detect CSS variables in form of var(--color) and get their current
+	 * values from the :root selector.
+	 */
+	if ( hex.indexOf( 'var(' ) > -1 ) {
+		let variableName = hex.match('--[a-zA-Z0-9]*')[0]
+
+		hex = window.getComputedStyle( document.documentElement )
+			.getPropertyValue( variableName ) || '#fff'
+
+		if ( hex.indexOf( 'rgb' ) > -1) {
+			const rgba = hex.replace( /^rgba?\(|\s+|\)$/g, '' ).split( ',' );
+
+			hex = `#${(
+				( 1 << 24 ) +
+				( parseInt( rgba[ 0 ] ) << 16 ) +
+				( parseInt( rgba[ 1 ] ) << 8 ) +
+				parseInt( rgba[ 2 ] )
+			)
+				.toString( 16 )
+				.slice( 1 )}`;
+		}
+	}
+
+
 	if ( ! alpha && 0 !== alpha ) {
 		return hex;
 	}


tomusborne:
Thanks for this!

Just tested, and I'm not sure it's working with GenerateBlocks as it is right now. Have you given it a shot (especially on the frontend)?

Can you expand a bit on what this PR is fixing? Is it allowing the Opacity slider to apply to `var()` values?

When I was playing with allowing global colors, I figured something like this would work fine:

```
if ( hex.indexOf( 'var(' ) > -1 ) {
    return hex;
}
```

As of right now, I'm not overly happy with having a separate value for opacity and having to depend on a function to convert it to RGBA - we'll be looking at deprecating the slider and simply allowed rgba values in the color picker.

Thanks again for the PR!

andreiglingeanu:
@tomusborne Completely agree. The other downside that I've found with the approach from this PR is the fact that when the global colors get updated, the change doesn't get picked up in the block settings, because it literally gets converted into a hex value at that point in time.

Deprecating the slider is not the best option too, frankly, because people may want to use the palette color (most often it's a solid color there) and apply just a slight opacity to it. Besides that, that's how the Gutenberg color picker works already, you'll have to make your own color picker to have this deprecation.

tomusborne:
Is it common to be able to add opacity to a CSS global variable color? I haven't seen much of that before - I feel like if a global color wants opacity, it should likely be defined that way?

I think the first step is to allow the picker to at least accept variables, which should be easy.

webhouse-dev:
Any updates on this?

tomusborne:
At this point, my plan is to:

1. Introduce an alpha slider inside the actual color picker, so it's not necessary to convert hex to rgba.
2. Allow variables as a value for the color picker.

So you won't be able to apply transparency to your global colors, but I think that's a good thing. You should apply the transparency to the global color itself when it's defined vs changing it in the UI.

andreiglingeanu:
@tomusborne Is there anything that we can do to help move this forward? Do you think we can get this solution merged in and get it to work on top of your solution? I don't see how these two would conflict even if both implemented. And in case you need to convert hex to rgba at some point -- variables will be handled out of the box.

---

Long term though, I don't like the conversion either. Because this way we loose the connection with the variable and you put the hex value at that point in time as the background. And this way further updates to the color palette will be ignored.

tomusborne:
I've actually done some work on this in the 1.4.0 branch. Our color picker no longer has an opacity slider, and it accepts CSS variables. I think it should handle most things this PR was hoping to fix. Let me know 🙂

andreiglingeanu:
@tomusborne confirmed, that branch handles everything correctly, thanks a lot! Closing this PR as it's no longer needed.

Hope this one's close to being released 🤞

PR #50, opened 2021-02-25
Always inline CSS on AMP pages to avoid needing to fetch

westonruter:
The AMP plugin inlines all CSS into the `style[amp-custom]` element. External stylesheets referenced by `link` elements get fetched and are then processed (minification and tree shaking). Having them be inline from the start avoids the need to fetch them from the filesystem (or worse) over HTTP.


tomusborne:
Thanks for this! Merged into our `release/1.3.2` branch that is being released next week. Appreciate it!

PR #61, opened 2021-06-04
Allow empty strings in Container padding

tomusborne:
Originally I believe the padding values were set to the `number` type which prevents them from being emptied. An empty string !== a number, so the default would always come back on editor reload. The solution was to force a `0` value if the field was emptied.

However, we switched to the `string` number type before release, and I'm not sure those forced `0` values are necessary anymore.


JeanPaiva (approved):
Looks good.

tomusborne:
Just updated this to merge into https://github.com/tomusborne/generateblocks/tree/try-default-removal.

JeanPaiva (approved):
The amount of deleted lines is so enjoyable. All good.

PR #63, opened 2021-06-09
Process widget content for block styling

tomusborne:
First try at this. It works, but I don't love it.

Some issues:

* It processes ALL widgets, even ones that don't appear on the current page. This means we're adding CSS to all pages for GB blocks that might not appear there.
* Performance impact?
* If the CSS is being served via an external file, that file needs to be regenerated every time a widget area is saved or the changes won't take place.

I think this opens the need for us to have a mechanism to always serve certain CSS inline vs. in the external file. CSS for widgets shouldn't be included in the external file.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -747,3 +747,72 @@ function generateblocks_get_svg_shapes() {
 		)
 	);
 }
+
+/**
+ * Loops through all registered widget areas and collects their data.
+ *
+ * @since 1.4.0
+ */
+function generateblocks_get_all_widget_data() {
+	global $wp_registered_sidebars;
+
+	$output = array();
+
+	foreach ( $wp_registered_sidebars as $sidebar ) {
+		if ( empty( $sidebar['name'] ) ) {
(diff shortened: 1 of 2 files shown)


tomusborne:
I think this is better now. One issue is that it uses `wp_get_sidebars_widgets()`, which is private: https://developer.wordpress.org/reference/functions/wp_get_sidebars_widgets/

I'm not sure why, exactly. I suppose we could extract the parts we need and ditch the function.

It still processes all widgets, even if they're not on the current page. I'm not sure a mechanism exists in WP to only target widgets on the current page. However, it will only add its CSS to an external file if GB blocks exist in the post content, so it should default to inline if that's not the case.

Performance shouldn't be an issue - looks like it's only one addition `get_option()` call.

Only way I could find to regenerate the CSS files on widget save was using the `widget_update_callback` filter. Seems to work.

JeanPaiva (approved):
It's way more cleaner. I want to test this with WP 5.8

PR #66, opened 2021-06-14
Style our blocks in the widget areas

tomusborne:
Original pull here: https://github.com/tomusborne/generateblocks/pull/63

Now going to merge this into 1.3.4 as a patch release for WP 5.8.

One issue with this PR is that it uses wp_get_sidebars_widgets(), which is private: https://developer.wordpress.org/reference/functions/wp_get_sidebars_widgets/

I'm not sure why, exactly. I suppose we could extract the parts we need and ditch the function.

### DIFF
--- a/includes/class-enqueue-css.php
+++ b/includes/class-enqueue-css.php
@@ -45,6 +45,7 @@ class GenerateBlocks_Enqueue_CSS {
 		add_action( 'save_post', array( $this, 'post_update_option' ), 10, 2 );
 		add_action( 'save_post_wp_block', array( $this, 'wp_block_update' ), 10, 2 );
 		add_action( 'init', array( $this, 'enqueue_assets' ) );
+		add_filter( 'widget_update_callback', array( $this, 'force_file_regen_on_widget_save' ) );
 	}
 
 	/**
@@ -434,6 +435,17 @@ class GenerateBlocks_Enqueue_CSS {
 	 */
 	public function update_saved_time() {
 		update_option( 'generateblocks_dynamic_css_time', time() );
+	}
+
+	/**
+	 * Force CSS files to regenerate after a widget has been saved.
+	 *
+	 * @param array $instance The current widget instance's settings.
+	 */
+	public function force_file_regen_on_widget_save( $instance ) {
+		update_option( 'generateblocks_dynamic_css_posts', array() );
+
+		return $instance;
 	}
 }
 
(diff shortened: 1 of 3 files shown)


JeanPaiva:
@tomusborne why not do something like this:

```php
function generateblocks_do_widget_styling( $content ) {
	$widget_blocks = get_option('widget_block');

	foreach ($widget_blocks as $key => $block) {
		if (preg_match('/(wp:generateblocks\/)/', $block['content'])) {
			$content .= $block['content'];
		}
	}

	return $content;
}
```

As far I saw WP saves all Widgets in `wp_options:widget_block`, the full html content. Other possibility is to hook after widget save and save the proper data the way we want.

tomusborne:
Is it really that easy? I spent so looking trying to find where they store the widget content.

JeanPaiva:
I think so, we have to run a lot of tests, but this pretty much did the job.

Here's GP:
Screenshot 2021-06-18 132622

Here's 2021 theme:
Screenshot 2021-06-18 132548

JeanPaiva (approved):
Tested on GP and 2021 theme. All good.

tomusborne:
Awesome, were you able to do any performance checks on `preg_match()` when a lot of widget  content exists?

JeanPaiva:
@tomusborne tested with 10 widgets here's the results:

Using preg_match:
preg_match

Using strpos:
strpos

The difference is not that big, and was tested in a controlled environment. `strpos` performed a bit better, I think we can switch to it. Not a huge gain, but one bit at a time.

tomusborne:
Awesome. I'm curious, what happens if we don't check at all and just feed all the content through it? Any difference?

JeanPaiva:
@tomusborne I did a couple more tests with 20 widgets:

`array_reduce` without checking the GB string:
reduce_without_check

`foreach` and `strpos`:
foreach_and_strpos

`foreach` without the check. It's the best result of all.
foreach_without_check

Here's the fastest solution:
```php
	$widget_blocks = get_option( 'widget_block' );

	foreach ( (array) $widget_blocks as $block ) {
		if ( isset( $block['content'] ) ) {
			$content .= $block['content'];
		}
	}

	return $content;
```

tomusborne:
That's as simple as it gets. Love it. Thanks for the performance tests!

JeanPaiva (approved):

PR #67, opened 2021-06-14
Prevent memory leak when reusable block is inside itself

tomusborne:
Right now there is a memory leak that happens if a user adds a reusable block inside of itself. This PR prevents that leak from happening by adding the reusable block ID into an array and checking against that array before parsing reusable blocks.


JeanPaiva:
@tomusborne could not test this issue properly. Not sure if I'm testing correctly, but there's something weird with the frontend even with in the branch. I think we need to sync, let me know.

JeanPaiva (approved):

PR #68, opened 2021-06-21
Fix block categories in WP 5.8

tomusborne:
In the WP 5.8 widget area, GB blocks are uncategorized as WP introduced a new filter and didn't add backwards-compatibility in the widget area (guessing): https://www.screencast.com/t/fiy3Ugzln

This should fix that issue while maintaining compatibility with versions prior to 5.8.

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -120,7 +120,11 @@ function generateblocks_do_block_editor_assets() {
 	);
 }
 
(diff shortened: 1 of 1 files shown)


JeanPaiva (approved):
Tested on 5.7.2 and 5.8 RC both working now.

Issue #70, opened 2021-07-07
Number attributes with defaults can't be cleared


tomusborne:
### Description
Right now, any `number` attribute with a default value can't be cleared to have no value. After saving, the default value simply comes back.

Right now, we fight this by adding a `0` value to these fields if they're emptied. This method works sometimes, but it:

1. Outputs CSS even if it's not necessary.
2. Doesn't allow Global Styles (GB Pro) to take over the local styling.
3. Doesn't allow you to set a grid item width to nothing.

### Steps to reproduce
Find a `number` type attribute (grid gaps, grid item width) and try to remove the default.

#### Expected behavior
The field should allow you to empty the default and it should remain empty.

#### Actual behavior
The field either doesn't allow you to empty it, or the default comes back on page reload.

#### Solution
We need to remove the defaults from these fields so they are able to be empty.

Simply removing the defaults will cause designs to change when users update, so we need to save the old defaults as static values when the user updates.

The following fields need to be changed. We'll handle them in different PRs.

- [x] Grid item widths - https://github.com/tomusborne/generateblocks/pull/92
- [x] Grid horizontal gap - https://github.com/tomusborne/generateblocks/pull/75
- [x] Grid vertical gap - https://github.com/tomusborne/generateblocks/pull/75
- [x] Gradients - https://github.com/tomusborne/generateblocks/pull/93

JeanPaiva:
@tomusborne I'll prioritize working on GP 3.1.0 for now. The PR's associated here I'll review them later this week.

tomusborne:
All of these are complete with the exception of the Container width option. I think that might be the one option that makes sense to have a default. However, I think we need to add an option the user can set for it in GenerateBlocks > Settings.

PR #71, opened 2021-07-07
Convert grid gap attributes to strings

tomusborne:
Related: https://github.com/tomusborne/generateblocks/issues/70

This converts the grid-gap options to the `string` type so we can clear them if needed.

Obviously, it's important that existing grid gap values our users have set are fully retained when they update. From my tests everything seems good to go, but it would be great to get some further testing.

1. Set up a page using master with a couple different grids.
2. Set one grid to use default horizontal gap, custom vertical gap.
3. Set another grid to use two custom gap values.
4. Update to this branch and check the front-end - all good?
5. Load the page in the editor - all good?
6. Save the page and view the front-end - all good?

This should get us closer to fixing this: https://community.generateblocks.com/t/global-styles-for-grids/66


tomusborne:
This is no longer needed now that we're removing defaults: https://github.com/tomusborne/generateblocks/pull/92

PR #72, opened 2021-07-07
Convert width attributes to strings

tomusborne:
Related https://github.com/tomusborne/generateblocks/issues/70, https://github.com/tomusborne/generateblocks/pull/71

This converts our grid item `width` attributes to `strings` so we can clear them easily.

Like the others, we should ensure that users retain their existing default and custom values when they update to this.

This PR also introduces a custom `RangeControl` component. The core component doesn't accept empty values (as it's built to work with `number` attributes). This one accepts empty values so it will work nicely with our `string` values.


tomusborne:
This is no longer needed now that we're removing defaults: https://github.com/tomusborne/generateblocks/pull/92

PR #75, opened 2021-07-14
Remove default attributes

tomusborne:
This is the first try at removing/replacing a block attribute default in a backward-compatible way.

Basically, we do the following:

1. Provide a function with our old default(s) we're replacing. This function returns through the defaults filter so we can account for filtered values and retain them.
2. In the editor, we check if the option has a value. If it doesn't, it means the user never saved an option and was using the default. So, we get the default from our function mentioned above and save it for them. This only runs once as we set `blockVersion` and check against it.
3. On the front-end, we hijack the setting with the defaults from our legacy defaults function. This only runs until the user saves the block (as `blockVersion` is then set).

This PR should:

1. Retain existing non-default values.
2. Set the previous default value as a static value when the block is loaded in the editor.
3. Use the previous default value on the front-end until the above happens
4. Not set the old default if the block is using a global style

If this system works, we can re-use it to remove default values from all of our blocks. This will close #71 and #72 as having an empty default will allow number attributes to be cleared.

If the logic makes sense and covers all angles we should look at cleaning it up so it's re-usable across multiple options.


JeanPaiva:
@tomusborne this is working fine if you have an old page and a new page, it's showing the 'horizontalGap' on the block settings. But if with my added filter to change the default it's not working as expected. I'll give it a try on this.

```php
add_filter('generateblocks_defaults', function($defaults) {
	$defaults['gridContainer']['horizontalGap'] = 20;

	return $defaults;
});
```

tomusborne:
If a user is setting a value like that using a filter, nothing should change for them. Our default change shouldn't affect them at all, so their blocks should continue to use that `20` default.

tomusborne:
Some scenarios we should account for:

1. User has added a block and kept the default values (changed nothing). Updating to this version should not change the default values - nothing should change for them. This PR should save the old default as a static value.
2. User has added a block and changed the default values (using static values). Updating should not affect their block.
3. User has used the `generateblocks_defaults` filter to add custom defaults and added a block using these new defaults. Updating should not affect their block.

JeanPaiva (approved):
All good

src/utils/set-block-version/index.js
```diff
@@ -0,0 +1,7 @@
+export default function setBlockVersion( props, blockVersion ) {
+	if ( 'undefined' === typeof props.attributes.blockVersion || props.attributes.blockVersion < blockVersion ) {
```
JeanPaiva, line 2: I think you can remove this function and do something like this:

```js
if ( isBlockVersionLessThan( props.attributes.blockVersion, 2 ) ) {
  setAttributes( { blockVersion: 2 } );
}
```

JeanPaiva (approved):

PR #77, opened 2021-07-15
Add js unit tests

JeanPaiva:
This PR adds JS unit tests and the JS CI action.

### DIFF
--- a/.github/workflows/js-ci.yml
+++ b/.github/workflows/js-ci.yml
@@ -1,0 +1,37 @@
+name: GenerateBlocks JS CI
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/**'
+  pull_request:
+    branches:
+      - master
+      - 'release/**'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check node_modules cache
+        id: npm-cache
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/package.json') }}
+          restore-keys: |
+            ${{ runner.os }}-modules-
+      - name: NPM install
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,28 +22,21 @@
 		"dompurify": "^2.2.8"
 	},
 	"devDependencies": {
-		"@wordpress/api-fetch": "^5.0.0",
```
tomusborne, line 25: Any reason to remove all of these? They weren't necessary before but it seems like `wp-scripts` changed at some point and now requires that we install each one we need.

package.json
```diff
@@ -22,28 +22,30 @@
...
+	  	"@wordpress/components": "^14.0.0",
+	  	"@wordpress/compose": "^4.0.0",
+	 	"@wordpress/data": "^5.0.0",
+	  	"@wordpress/element": "^3.0.0",
+	 	"@wordpress/hooks": "^3.0.0",
+	  	"@wordpress/i18n": "^4.0.0",
+	 	"@wordpress/rich-text": "^4.0.0",
+	  	"@wordpress/scripts": "^16.0.0",
+		"enzyme": "^3.11.0",
+		"grunt": "^1.2.1",
```
tomusborne, line 37: Did we need to downgrade this from 2.0.0? Same with other grunt packages?

JeanPaiva: Mine rebase messed up. I started this branch from master and switched later. I'll re-do my rebase and should fix all these issues.

tomusborne (approved):

.github/workflows/js-ci.yml
```diff
@@ -0,0 +1,37 @@
+name: GenerateBlocks JS CI
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/**'
```
tomusborne, line 7: GitHub is no longer letting me put `/` in the branch name - it gets converted to `-`.

PR #80, opened 2021-07-16
wp-editor should not be enqueued in the new widget editor

### LINKED ISSUE #78: wp-editor should not be enqueued in the new widget editor
### Description
In WP 5.8-RC3, we're seeing this notice: https://www.screencast.com/t/pZst8jRoE87

This is due to this change in core: https://github.com/WordPress/wordpress-develop/pull/1484

I'm not 100% sure if this dependency is needed in the editor, but we need to remove the `wp-editor` dependency (https://github.com/tomusborne/generateblocks/blob/1.3.4/includes/general.php#L26) when inside the block widget editor.

### Steps to reproduce
Open the block widget editor in 5.8 when GB is activated.

#### References: 
https://github.com/WordPress/wordpress-develop/pull/1484
https://core.trac.wordpress.org/ticket/53569

JeanPaiva:
Closes #78

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -20,10 +20,18 @@ add_action( 'enqueue_block_editor_assets', 'generateblocks_do_block_editor_asset
  * @since 0.1
  */
 function generateblocks_do_block_editor_assets() {
+	global $pagenow;
+
+	$generateblocks_deps = array( 'wp-blocks', 'wp-i18n', 'wp-editor', 'wp-element', 'wp-compose', 'wp-data' );
+
+	if ( 'widgets.php' === $pagenow ) {
+		unset( $generateblocks_deps[2] );
+	}
+
 	wp_enqueue_script(
 		'generateblocks',
 		GENERATEBLOCKS_DIR_URL . 'dist/blocks.js',
-		array( 'wp-blocks', 'wp-i18n', 'wp-element', 'wp-editor', 'wp-compose', 'wp-data' ),
+		$generateblocks_deps,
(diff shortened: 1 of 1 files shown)


includes/general.php
```diff
@@ -20,10 +20,18 @@
  * @since 0.1
  */
 function generateblocks_do_block_editor_assets() {
+	global $pagenow;
+
+	$generateblocks_deps = array( 'wp-blocks', 'wp-i18n', 'wp-editor', 'wp-element', 'wp-compose', 'wp-data' );
+
+	if ( wp_use_widgets_block_editor() && 'widgets.php' === $pagenow ) {
```
tomusborne, line 27: We should use `function_exists()` here to prevent a fatal on WP < 5.8. We may actually be able to get away without it at all as `enqueue_block_editor_assets` should only fire on pages with the block editor, so the `$pagenow` check may be enough.

includes/general.php
```diff
@@ -20,10 +20,18 @@
...
+		unset( $generateblocks_deps[2] );
```
tomusborne, line 28: Need to remember to keep the order here if we ever add/remove items. In a bigger release we should review whether we even need these dependencies as we `import` them in the JS files.

includes/general.php
JeanPaiva: Removed the function check.

includes/general.php
JeanPaiva: Yes, but we are going to possibly remove all these dependencies since in 1.4.0 we import them via NPM.

tomusborne (approved):

PR #91, opened 2021-08-13
Remove container padding defaults

tomusborne:
This removes the 40px container padding default.

This one has definitely been filtered by users (https://community.generateblocks.com/t/change-the-default-container-padding/287) so it's important we check the implications of this on filtered defaults (different values or no value at all).


src/blocks/container/edit.js
```diff
@@ -128,13 +129,28 @@ class GenerateBlockContainer extends Component {
 			}
 		}
 
-		// Update block version flag if it's out of date.
-		const blockVersion = 2;
+		// Set our old defaults as static values.
+		// @since 1.4.0.
+		if ( isBlockVersionLessThan( this.props.attributes.blockVersion, 2 ) ) {
```
tomusborne, line 134: One issue here is that this returns true if it's a fresh new block as the `blockVersion` isn't set yet. This means new Containers (in this case) will add the old default as static values by default.

JeanPaiva: @tomusborne adding `blockVersion` with a default value of two in the `container/attributes.js` would not solve this problem?

tomusborne: Then the block would always use the default and never save a static version value. That would make it so we'd never know if a block had run any migration code (less than would never be true).

tomusborne: This should fix it: https://github.com/tomusborne/generateblocks/pull/91/commits/89cd6ee9d94e0dcfd2b0903a50b13a8517592c85

JeanPaiva (approved):
All good.

PR #92, opened 2021-08-13
Remove grid item width defaults

tomusborne:
This removes the default width and mobile width defaults from our grid item containers.

The only downside to doing this on mobile is we have to repeat the `width: 100%` CSS for each grid item on mobile if it's set that way. Previously we only had to print it once as it was the default.


JeanPaiva (approved):
All good, working as expected.

PR #93, opened 2021-08-13
Remove container and button gradient default values

JeanPaiva:
This removes gradient default values for container and button blocks.


includes/class-legacy-attributes.php
```diff
@@ -72,6 +84,23 @@ public static function get_settings( $version, $block, $settings, $atts ) {
 				$settings['paddingRight'] = $legacy_settings['paddingRight'];
 				$settings['paddingBottom'] = $legacy_settings['paddingBottom'];
 				$settings['paddingLeft'] = $legacy_settings['paddingLeft'];
+
```
tomusborne, line 87: Missing `gradientDirection`

includes/class-legacy-attributes.php
```diff
@@ -72,6 +84,23 @@ public static function get_settings( $version, $block, $settings, $atts ) {
...
+				$settings['gradientColorTwo'] = $legacy_settings['gradientColorTwo'];
+				$settings['gradientColorTwoOpacity'] = $legacy_settings['gradientColorTwoOpacity'];
+			}
+
+			if ( 'button' === $block ) {
+				$button_legacy_settings = wp_parse_args(
+					$atts,
+					$legacy_defaults['button']
+				);
+
```
tomusborne, line 99: Missing `gradientDirection`

tomusborne (commented):
Also missing the code that generates the legacy CSS on the front-end for buttons. For example: https://github.com/tomusborne/generateblocks/blob/try-default-removal/includes/generate-css.php#L86-L91

tomusborne (approved):
This looks good. We should keep in mind that we're using `hasNumericValue()` for non-numeric values (in other places too). We should probably add a `hasValue()` util soon.

JeanPaiva:
The function itself is a value check we can rename all occurrences to `hasValue`. In the future, we may want to add a `getNumberValue` function which parses the value to an integer, check if is NaN, and return the formatted value.

PR #94, opened 2021-08-15
Remove desktop autoWidth

tomusborne:
This removes the "Auto" option from the grid item width on desktop.

This is no longer needed once we remove the default value (https://github.com/tomusborne/generateblocks/pull/92), as the user can simply clear the width value to achieve an "auto" width.

We've left the "Auto" option on tablet and mobile as those actually output an `auto` property value to override the parent.


JeanPaiva:
We might want to keep this PR, in #92 branch I couldn't clear the width value properly.

tomusborne:
Working on a PR for that today.

tomusborne:
@JeanPaiva this good to go now?

JeanPaiva (approved):
All good.

PR #96, opened 2021-08-15
Remove select grid button

tomusborne:
This removes the Grid button from the toolbar in grid item Containers. It's no longer needed as WP added the parent block button in the toolbar in 5.8.


JeanPaiva (approved):
All good. You need to wait for try-remove-defaults to be merged or cherry-pick the commit that we disabled the CI (2672ec8)

PR #98, opened 2021-08-17
Add custom range control

tomusborne:
This adds a custom range control (taken from GB Pro).

The differences between this control and the core control:

1. It can be empty (no value)
2. You can control range step/min/max separately to the input

This also changes the grid item width control to use this control so we can allow for empty values.


src/components/range-control/index.js
```diff
@@ -0,0 +1,85 @@
...
+import {
+	Component,
+} from '@wordpress/element';
+
+import {
+	RangeControl,
+	TextControl,
+} from '@wordpress/components';
+
+export default class RangeControlInput extends Component {
```
JeanPaiva, line 15: Geat job here. Simple and powerful. What about adding some unit tests for this component?

JeanPaiva (approved):
All good. The grid container slider feels way more slick and useful with new values, really great job.

src/components/range-control/index.js
tomusborne: Would be awesome!

PR #99, opened 2021-08-18
Fix wasBlockJustInserted when adding grid items

### LINKED ISSUE #97: wasBlockJustInserted fails when adding grid items
### Description
In #75 we're using `wasBlockJustInserted` to make sure newly added blocks don't run the migration and get the old defaults as static values.

This fails when using the Grid block. When you choose a selected layout, only the last block inserted returns true for `wasBlockJustInserted`, making it so the previous grid items have those old defaults saved as static values.

### Steps to reproduce
1. Using #75, add a Grid block
2. Choose a layout with more than one column

#### Expected behavior
All grid items should look the same.

#### Actual behavior
All grid items except for the last one have our old 40px of padding.

tomusborne:
This fixes #97.

It adds a `wasBlockJustInserted` attribute to the Container block set to `false` by default.

When adding grid items using the Grid block, it sets this attribute to `true`.

Once the Container block has mounted, we set it back to its default so the attribute doesn't remain in the comment delimiter.


src/blocks/container/edit.js
```diff
@@ -133,7 +132,7 @@ class GenerateBlockContainer extends Component {
 
 		// Set our old defaults as static values.
 		// @since 1.4.0.
-		if ( ! wasJustInserted && isBlockVersionLessThan( this.props.attributes.blockVersion, 2 ) ) {
+		if ( ! wasBlockJustInserted && isBlockVersionLessThan( this.props.attributes.blockVersion, 2 ) ) {
```
JeanPaiva, line 135: This should not be called as `wasBlockJustInserted( props )` to get the correct value? Cause you imported the function, and are not using it from the props the way you were doing before.

src/blocks/container/edit.js
```diff
@@ -163,6 +162,10 @@ class GenerateBlockContainer extends Component {
 			}
 		}
 
+		if ( wasBlockJustInserted ) {
```
JeanPaiva, line 165: Same as above.

JeanPaiva (commented):
The tests are working as expected.

src/blocks/container/edit.js
tomusborne: Nice catch.. Now when I do this, the `wasBlockJustInserted` prop is `false` even when freshly inserting it.. Wonder why it stopped working all of a sudden.

JeanPaiva: But this check is actually being always true, cause it checks if `wasBlockJustInserted` is defined. I'll check this behavior to see if I can catch something

tomusborne: Yea, that's definitely a bug.

What's weird is that even when fixing that, `select( 'core/block-editor' ).wasBlockJustInserted( props.clientId )` is now `false` even when inserting it as a fresh block. Not sure why that stopped working..

tomusborne: Hmm, seems to be working now... Let me know how it works for you.

JeanPaiva (approved):
It's working as expected.

PR #103, opened 2021-08-23
Fix gradient static values always set

### LINKED ISSUE #102: Static gradient values saving when not needed
### Description
In #75 we've removed gradient defaults and have made it so the old defaults are added as a static values once the block is loaded in the editor.

However, we don't check whether gradients are active or not before we do this. This will make it so blocks that aren't using gradients add a bunch of values without actually needing to: https://www.screencast.com/t/wn43IoaoPxtb

Since we set these values when "Use Gradient" is toggled, I don't think we need to set them if it isn't toggled.

JeanPaiva:
closes #102 

**Gradient not working on the frontend, I'm still working to fix it.**


tomusborne:
We also need to check if gradients are being used when running the migration: https://github.com/tomusborne/generateblocks/blob/try-default-removal/src/blocks/container/edit.js#L148-L152

Otherwise our migration code will save static gradient values for users who aren't even using gradients.

JeanPaiva:
@tomusborne I fixed what you mentioned above and also had to do some quick fixes on PHP to properly generate the CSS.

includes/functions.php
```diff
@@ -365,7 +365,11 @@ function generateblocks_hex2rgba( $hex, $alpha ) {
 		return;
 	}
 
-	if ( 1 === $alpha ) {
+	if ( strpos($hex, 'rgba') !== false ) {
```
tomusborne, line 368: What did this fix? Might be worth putting it under the next check to prevent `strpos()` from firing at all if `$alpha` is `1` or `''`.

JeanPaiva: We now have values that are already saved in `rgba` if we match rgba we bail the `$hex`

tomusborne: Shouldn't `$alpha` always be a number or `''`? Did you find a case where it was something else?

JeanPaiva: @tomusborne I removed this `strpos` check, I re-tested and it works only checking for empty string.

includes/functions.php
```diff
@@ -365,7 +365,7 @@ function generateblocks_hex2rgba( $hex, $alpha ) {
 		return;
 	}
 
-	if ( 1 === $alpha ) {
+	if ( 1 === $alpha || '' === $alpha ) {
```
tomusborne, line 368: If we want to be extra cautious, we could consider something like this:

```
if ( 1 === $alpha || ! is_numeric( $alpha ) ) {

}
```

That would account for `''` or `false` etc.. Thinking about Advanced Gradients in GB Pro where it might not have a default.

JeanPaiva: This is a good idea. Pushed the change.

tomusborne:
I'm still seeing the issue where toggling the gradient option removes options set by the user: https://www.screencast.com/t/NEsEPh8nTd

Are you able to replicate that?

Also found out why you did the `rgba()` check while messing around - it completely breaks if you're using rgba colors and the migration happens to add opacity. Not sure it's an issue though as users shouldn't be able to use rgba colors in the current version.

PR #106, opened 2021-08-27
Allow zero values for minHeight on devices

tomusborne:
This allows for `0` as a value when setting `minHeight` on tablet and mobile.


JeanPaiva (approved):
All good.

PR #107, opened 2021-08-31
Update changelog for 1.4.0

tomusborne:
First draft of changelog. Need to review the commits again to make sure nothing else needs to be added.


JeanPaiva:
We need to update this with the latest PR's we opened.

tomusborne:
Most of those are dealing with new features we've already mentioned in the changelog. Think this is good to go.

PR #108, opened 2021-09-01
Disable grid item width options when using flex-basis

tomusborne:
This PR disables the width field/slider if we're using the Flex Basis option.

It's a bit of a logic minefield..

We want to:

1. Disable the width options across all lower devices if flex basis is set
2. Allow the width option on those lower devices if flex-basis is set manually to "auto"

See discussion here for more context: https://edge22studios.slack.com/archives/C023546D3QR/p1630528829016900?thread_ts=1630512448.010400&cid=C023546D3QR


JeanPaiva:
It works as expected, but I can add anything to `basis` field. Maybe we can change to a select field?
Screenshot 2021-09-02 193404

tomusborne:
Hmm, it can be any percentage or "auto". Maybe it just needs some basic sanitizing to allow numbers, decimals and lowercase characters?

JeanPaiva:
Ye could help, just to not be so open.

tomusborne:
Visual validation looks pretty complex - surprised it's not a core thing. For now I think the above should do the trick. We should look into adding our own validation method that's better for users in the future.

JeanPaiva (approved):
It works as expected. If the user does a typo it will remove the full string. We may need to improve the input field in the future to allow better feedback for users.

PR #109, opened 2021-09-02
Check for empty uniqueId in wasBlockJustInserted

tomusborne:
Checking for the `wasBlockJustInserted` prop was unreliable, especially when it came to replacing blocks (like the initial paragraph block when you create a new page).

Instead, this checks to see if `uniqueId` is an empty string. 

`uniqueId` is set previously in `componentDidMount`, but it doesn't seem to matter during the check lower down.


JeanPaiva (approved):
All good.

tomusborne:
Do you prefer this to using state?

JeanPaiva:
Two different ways of doing the same thing, since this is working and we have the alpha release on Monday, we can go with your code, perfectly fine.

PR #110, opened 2021-09-06
Add reset button to flex options

tomusborne:
This adds a reset button to our flex options if one of them has a value.

This allows for a one-click reset instead of having to potentially clear three separate inputs.

It adds a `gblocks-utility-label` element we can reuse later if we want. Might be worth looking at merging something like this with the `UnitPicker` component.


src/blocks/container/edit.js
```diff
@@ -636,9 +637,32 @@ class GenerateBlockContainer extends Component {
...
 									>
+										<div className="gblocks-utility-label">
+											<label
+												htmlFor="gblocks-flex-grow-desktop"
+												className="components-base-control__label"
+											>
+												{ __( 'Flex', 'generateblocks' ) }
+											</label>
+
+											{ ( hasValue( flexGrow ) || hasValue( flexShrink ) || hasValue( flexBasis ) ) &&
```
JeanPaiva, line 649: I don't think it's necessary to check for values to display the icon, the button can always be displayed in my opinion.

src/blocks/container/edit.js
```diff
@@ -636,9 +637,32 @@ class GenerateBlockContainer extends Component {
...
+												<Tooltip text={ __( 'Reset', 'generateblocks' ) } position="top">
+													<Button
+														className="gblocks-inline-reset-button"
+														icon={ getIcon( 'refresh' ) }
```
JeanPaiva, line 653: I don't think the icon matches exactly the intent it seems you can go back to old values. Maybe changing to a `delete` (backspace) icon of Feather icons. This is totally opinionated though.

src/blocks/container/edit.js
```diff
@@ -825,9 +849,32 @@ class GenerateBlockContainer extends Component {
 
 									<BaseControl
 										className="gblocks-flex-controls"
-										label={ __( 'Flex', 'generateblocks' ) }
-										id="gblocks-flex-grow-tablet"
 									>
+										<div className="gblocks-utility-label">
```
JeanPaiva, line 853: This hole block should be moved to a component, later on, to re-use logic and avoid code duplication.

JeanPaiva (commented):
Testing is ok tho.

src/blocks/container/edit.js
tomusborne: Agreed, was getting a little fancy haha

src/blocks/container/edit.js
tomusborne: What are other plugins doing? Wonder if Gutenberg itself is using an "undo" or "clear" icon.

src/blocks/container/edit.js
tomusborne: Agreed - it should be merged with the `UnitPicker` component.

src/blocks/container/edit.js
JeanPaiva: I've no idea, but we can leave it like this for now, if users complaining about having confusion we can change it.

tomusborne: Better or worse now?

JeanPaiva: Yes, it's better IMO. 👍

JeanPaiva (approved):
All good.

PR #111, opened 2021-09-06
Prevent grid item zero values or values that start with zero

tomusborne:
This prevents users from setting a `0%` width for grid items.

It also tightens up the UI a little by placing the slider inside the `BaseControl`.


JeanPaiva (approved):
All good.
Quick note: Users are still allowed to use fractional numbers. Eg. `.2`

tomusborne:
Hmm, think we should prevent that in the `startsWith` check as well? Might be a good idea.

JeanPaiva:
@tomusborne but doing so are we not deciding for the user that it should not use these small fractions of size? Not sure of a use case for this, but do we have issues allowing them to use this?

tomusborne:
Can't think of a good reason not to allow it. Not allowing 0 makes sense, but if a user needs to set `.1` then they can.

PR #113, opened 2021-09-06
Add new block on Enter key in Headline block

tomusborne:
This makes it so hitting the "Enter" key while in the Headline block creates a new Paragraph block underneath.

Currently, it makes the Headline block multiline, which is expected to happen when hitting "Shift + Enter".

It also accounts for hitting "Enter" while in the middle of some Headline text. It will then split the Headline into two separate Headline blocks.

I'm not exactly sure where `isOriginal` comes into play. This implementation mirrors the core Heading block: https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/heading/edit.js#L62-L79

I dug into core to see if `isOriginal` is needed, but couldn't find anything.


src/blocks/headline/edit.js
```diff
@@ -238,6 +244,25 @@ class GenerateBlockHeadline extends Component {
 
 		htmlAttributes = applyFilters( 'generateblocks.frontend.htmlAttributes', htmlAttributes, 'generateblocks/headline', attributes );
 
+		const onSplit = ( value, isOriginal ) => {
```
JeanPaiva, line 247: The split logic it feels strange on my understanding. If I want to finish a word and break a line I go back to a normal paragraph but if I wanted to stay using a Headline? I would have to add 3-4 extra actions for me to bring back the headline. It does not feel natural to me, not sure about this change.

tomusborne: Without this, it's not possible to add a new block under the Headline block without clicking the + and choosing a new block: https://www.screencast.com/t/UZMm6pTo

With this, hitting Enter adds a new paragraph block, which is what most people want after writing a heading: https://www.screencast.com/t/TF01RG2uiR

I'm not sure the splitting mid-Headline is necessary, can't see where that would be helpful really, but it is how the core Heading block works as well. Basically this change makes it so the Headline block behaves like the core Heading block. The not starting a new paragraph on Enter thing has tripped me up personally a few times while writing content.

JeanPaiva (approved):

src/blocks/headline/edit.js
tomusborne: This explains the `isOriginal` thing: https://github.com/WordPress/gutenberg/pull/28505

Basically, `onSplit` replaces both blocks (which explains why `onSplit` requires `onReplace`). So in this case it replaces the current Headline with another Headline with the same value/attributes and it creates an empty paragraph as the next block. Then it reassigns the `clientId` to the original block so it doesn't change.

tomusborne:
@JeanPaiva Happy with this, or still hesitant?

JeanPaiva:
I mean functionally it's working and the code is good. The thing is that I was thinking that the Headline could be used as a paragraph since you can set the `p` tag. My idea was to keep the user using our Block instead of the core one. But what you said about writing the Headline and jump right into the paragraph makes sense.

tomusborne:
That's a great point about the `p` tag - I think we can work that in later if needed/requested. Makes complete sense.

PR #116, opened 2021-09-08
Set grid item width when using grid toolbar button

tomusborne:
This gives our grid items a width when we use the + button in the Grid block toolbar. Right now it adds a Container with no width at all.


JeanPaiva (approved):
All good.

PR #119, opened 2021-09-10
Disable default container appender

tomusborne:
This changes `undefined` to `false` so we disable the core (paragraph) appender when inner blocks are selected.


JeanPaiva (approved):
All good.

PR #120, opened 2021-09-10
Fix error when using grid item width slider

tomusborne:
This fixes a `value.startsWith()` is not a function error when using the Grid item width slider.

`startsWith()` requires a string to work.


JeanPaiva (approved):
All good.

PR #121, opened 2021-09-14
Improve empty container appender

tomusborne:
This adds the tinycolor package, so be sure to install packages when testing.

This should fix [an issue](https://edge22studios.slack.com/archives/C022YLCJXHS/p1631310185114200) where the drop zone for blocks was limited to the + icon button when dropping them into an empty container.

This should also improve the overall UI/UX of the Container appender.

It adds the regular button appender when the Container is selected allowing the user to easily add more blocks.

It also changes the color of the appender based on the background color of the Container, so it doesn't get lost in some designs.

appender


JeanPaiva:
@tomusborne I'm getting this misalignment: (I'm on Firefox)

Screenshot 2021-09-14 152652

tomusborne:
Strange, here's what I'm seeing in Firefox:

appender

Can you send me the code editor code for that page so I can take a look?

tomusborne:
Should be fixed now.

JeanPaiva (approved):
All good.

PR #122, opened 2021-09-14
Fix container tag name accepting any value

JeanPaiva:
This adds a tag name attribute validation allowing only the defined tags.


tomusborne:
Works well! Seeing a couple of JSLint issues though.

jslint

JeanPaiva:
@tomusborne fixed!

tomusborne:
One issue I found is GB Pro doesn't actually filter the list of available tagnames - `a` isn't an option to choose from, the plugin does it for you if you enable container links.

So checking the list of tag names failed when it was set to `a`, as it's not an option in the select.

Last commit I did there was to introduce a new list of allowed tag names just like we have on the frontend. Might be a better way? Not sure.

JeanPaiva:
@tomusborne I liked. Just removed the `map` function and I think we are good to go now!

tomusborne (approved):

PR #123, opened 2021-09-15
Check for opacity value before clearing

tomusborne:
This fixes an undefined function error when clicking the "Clear" button in the color picker when the legacy opacity field isn't set.

undefined-function


JeanPaiva (approved):
All good.

PR #124, opened 2021-09-15
Add background color to appender

tomusborne:
Previously (https://github.com/tomusborne/generateblocks/pull/121) we tried to change the appender colors based on the Container background/text. It works, but is prone to failures (background images, container hover colors etc..).

Instead, I think it's better that we simply add a background color to the appender. This way users will never have to worry about not being able to see the appender.


JeanPaiva (approved):
All good.

PR #127, opened 2021-09-23
Add hash to hex values missing it

tomusborne:
Right now if you type or paste in a hex value into the color picker without the has (`000000`) it won't work, as the `#` is missing from the value.

This automatically adds the `#` if it's missing.

If this works nicely, we need to add it to the GBP and GP color picker components as well.


JeanPaiva (approved):
All good.

PR #128, opened 2021-09-25
Implement withUniqueId HOC component

JeanPaiva:
This PR refactor the logic to apply the attribute `uniqueId` to GB blocks. Is expected to keep the same behavior as before.


tomusborne (approved):
This looks good (love the tests).

My tests are showing a new ID being generated when a block is duplicated.

I'm also not seeing any existing IDs changing when switching to this PR.

One thing we should keep in mind for 1.5.0 is that `getBlocks()` doesn't work in the widget editor: https://github.com/WordPress/gutenberg/issues/32607

We should look at adding compatibility with the widget editor (or figuring out a different way to know all the unique IDs being used).

JeanPaiva:
Opened #129 to fix the widget blocks issue.

PR #130, opened 2021-09-28
Split headline component

JeanPaiva:
This PR only splits the BlockControls and InspectorControls components to separate files to reduce the size of the file. All behavior should stay the same.


tomusborne (approved):
Much nicer. Tested and not seeing any issues. All good.

PR #132, opened 2021-10-04
Split Grid block components

JeanPaiva:
This splits some internal components of the grid block to reduce the file size and increase readability.


tomusborne (approved):
Looks good!

PR #134, opened 2021-10-05
Improve empty container visibility

tomusborne:
Right now the empty container dashed border inherits the text color from the Container.

The problem here is that it requires that text color to be set to something visible in order to see the borders of the empty Container.

To test this, use `release/1.4.0` branch and add:

1. Container
2. Grid
3. Three columns

With a white background, everything looks fine. Now make the main container background black - the borders are no longer visible.

This PR should fix that issue allowing the empty container border to always be visible regardless of the container/content background.


tomusborne:
@JeanPaiva Did that a few different ways. Unfortunately it gets difficult because:

1. The current Container can have no `backgroundColor` value and still have a background because of its parent Container
2. We aren't aware of the editor background color if no Container background color is set

I know we could do some `getComputedStyle` stuff but I figure this is a decent solution for now as we're approaching release candidates.

PR #136, opened 2021-10-07
Fix align support in block themes

### LINKED ISSUE #87: Align wide/full missing when using theme.json
### Description
It's been reported that themes using `theme.json` do not display the align wide/full option in the Container block.

It seems block themes don't use `add_theme_support`, which GB depends on checking to add this option: https://fullsiteediting.com/lessons/global-styles/#h-layout-and-content-width

The reason we check for support and add it instead of using the built-in way of adding it is because we don't want Container blocks inside the Grid to have the option.

### Steps to reproduce
1. Add a block theme
2. Check the Container block for full/wide alignment options.

#### Expected behavior
The option should appear if the block theme supports it.

#### Actual behavior
The option doesn't appear.

#### References: 
- https://wordpress.org/support/topic/container-alignment-option-missing-if-theme-json-exists/#post-14726447

tomusborne:
Closes #87 

Previously we were checking `get_theme_support( 'align-wide' )` to see if we should include the wide/full alignment options in the Container block.

However, block-based themes don't use `add_theme_support()` (https://fullsiteediting.com/lessons/global-styles/#h-layout-and-content-width), so this check was failing for them.

This PR ditches the `get_theme_support()` check and uses `getBlockSupport()` (https://github.com/WordPress/gutenberg/blob/1da217e8a57803e2c9183316cdef3d09c618ffcb/packages/blocks/src/store/selectors.js#L264) with an empty block name and `true` as the default value. This seems to allow us to artificially set support for `align` to true only when we need it.

It should:

1. Show the alignment option in the toolbar for "classic" themes using `add_theme_support()` (GeneratePress)
2. Show the alignment option in the toolbar for block-based themes (https://wordpress.org/themes/tt1-blocks/)
3. Not show the alignment option for "classic" themes if `add_theme_support( 'align-wide' )` is not set
4. Not show the alignment option for block-based themes if the `layout` options aren't set in theme.json: https://github.com/WordPress/theme-experiments/blob/master/tt1-blocks/theme.json#L111-L114

It would obviously be easier if we could set the `align` supports option to true in block registration, but then we would:

1. Have wide/full alignment for grid items which isn't possible
2. Not have control over where in the toolbar the option appears


bradley2083:
What was the solution to this?  I'm also running into an issue when I have theme.json and add_theme_support( 'align-wide' ); with the align wide and full options not appearing in the editor.

markdavies:
I'm still seeing this issue as well. `add_theme_support( 'align-wide' );` has no effect for image blocks

flmuel:
The theme.json overrides functions added via add_theme_support(). This is known and ok. But strangely the theme.json overwrites also the function "align-wide" although this is not defined in the schema or can't be configured there at all.

This is not fixed :(

flmuel:
The problem lies in the merging of the settings. The error is in the function "get_theme_data()" in the "WP_Theme_Resolver_Klasse" -> https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-theme-json-resolver.php#L195 where "Theme Support" settings is mapped to the SCHEMA. I.e. "align-wide" and possibly other features are simply removed here.

Zsolt-R:
This problem reappeared in version 15.1.0 
With a clean installation, created a test child theme containing:
- style.css 
- index.php 
- functions.php with `add_theme_support( 'align-wide' );`

Create a new page and insert a cover block and you have the align wide/full controls, then add a basic theme.json with only this content:
`{
  "$schema": "https://schemas.wp.org/trunk/theme.json",
  "version": 2,
  "settings": {
    "layout": {
      "contentSize": "650px",
      "wideSize": "1200px"
    }
  }
}`
the align wide/full disappears.

strarsis:
+1, same issue again with latest Gutenberg!

PR #139, opened 2021-10-13
Update color palette on input

tomusborne:
Right now when you manually type/paste into the color picker input, it doesn't update the palette up above until `onBlur`: https://www.screencast.com/t/GAlMft9dv

This sets a timer so it updates on input: https://www.screencast.com/t/s8l27lMHQ6

I think we can do this "nicer" if it was a functional component, but this seems to work. If it's good, we'll want to do the same in the GP/GBP color picker components.


src/components/color-picker/index.js
```diff
@@ -141,11 +143,16 @@ export default class GenerateBlocksColorPicker extends Component {
 											}
 
 											onChange( color );
-										} }
-										onBlur={ () => {
-											this.setState( {
-												colorKey: value,
-											} );
+
+											clearTimeout( this.timer );
```
JeanPaiva, line 147: If you type and quickly click outside so the popover disappears it throws an error.

```
Uncaught TypeError: document.querySelector(...) is null
```

This can be fixed by clearing the timeout on unmount:
```js
componentWillUnmount() {
     clearTimeout( this.timer );
}
  ```

tomusborne: Nice catch! Also had to check for the input to focus as doing the same thing caused an error. Should be fixed now.

PR #154, opened 2021-10-30
Filter the font list of the typography component.

loxK:
Add a filter, allowing to filter the font picker's fonts. This allows theme developers to add their own local fonts.

See: https://generatepress.com/forums/topic/add-local-font-to-generateblock-dropdown

### DIFF
--- a/src/components/typography/index.js
+++ b/src/components/typography/index.js
@@ -26,6 +26,10 @@ import {
 	Button,
 } from '@wordpress/components';
 
+import {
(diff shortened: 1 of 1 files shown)


JeanPaiva:
@loxK thank you for your PR. This is the actual code needed for this issue at the current state of the code.
Internally we are discussing some improvements that we want to do for the Typography component in general, for now, we are not going to merge this as we might end up changing the component entirely in a near future.

PR #165, opened 2021-11-15
Adds query loop functionality to the grid block

JeanPaiva:
This PR adds the base functionality for the query loop. It should be easier to extend this structure to add more control over the query.


src/blocks/container/block-controls.js
```diff
@@ -75,7 +77,7 @@ const withAdvancedControls = createHigherOrderComponent( ( BlockEdit ) => {
 
 		return (
 			<Fragment>
-				{ isGrid && parentGridId &&
+				{ ! isQueryLoop && isGrid && parentGridId && 'generateblocks/container' === name &&
```
tomusborne, line 80: I think we removed the name check to prevent this button from disappearing in the Grid block.

src/blocks/grid/components/QueryLoopRenderer.js
```diff
@@ -0,0 +1,76 @@
...
+			<BlockContextProvider
+				key={ postContext.postId }
+				value={ postContext }
+			>
+				{ postContext === ( activeContext || dataContexts[ 0 ] )
+					? <InnerBlocks { ...props } templateLock={ templateLock }  />
+					: <div className={'block-editor-inner-blocks'}>
+						<BlockPreview
+							blocks={ innerBlocks }
+							__experimentalLive
```
tomusborne, line 69: What happens if they remove/rename the `__expiremental` features? Will it gracefully fallback or break things completely?

src/blocks/grid/attributes.js
```diff
@@ -12,6 +12,9 @@ export default {
 		type: 'number',
 		default: '',
 	},
+	isQueryLoop: {
```
tomusborne, line 15: Should we give this a `default` of `false` to prevent it from being added to the block settings unnecessarily?

src/blocks/grid/components/QueryLoopRenderer.js
JeanPaiva: It won't break the component, but we are going to be seeing the first post only, we lose the preview for all posts currently showing.

JeanPaiva: The core query loop is using both of these props. Experimental stuff I think we just need to take care of once we update npm packages.

tomusborne: Cool, as long as things don't explode. We'll just need to keep an eye on WP development for when these things are given a stable name.

src/blocks/grid/components/query-loop/InspectorControls.js
```diff
@@ -0,0 +1,32 @@
...
+	const [ isQueryLoop, setIsQueryLoop ] = useState( attributes.isQueryLoop );
+	const { replaceInnerBlocks } = useDispatch( blockEditorStore );
+
+	useEffect( () => {
+		setAttributes( { isQueryLoop } );
+		replaceInnerBlocks( clientId, [] );
+	}, [ isQueryLoop ] );
+
+	return (
+		<InspectorControls>
```
tomusborne, line 18: I think this should probably show up near the bottom - above Documentation and Advanced.

Maybe we should filter the Documentation section in so we can easily add things above it? Like we do with the Container block controls: https://github.com/tomusborne/generateblocks/blob/release/1.5.0/src/blocks/container/block-controls.js

I can tackle that if you think it's necessary?

JeanPaiva: I added at the top only for test visibility, we can change position at any time. I opened #166 to make 'ResposiveTabs' a HOC so we could place the query loop right below.

My idea for all these controls was to make use of `compose` to display the correct orders. But we can make use of filters for simplicity on GBP.

Simple example:
```js
compose(
    withResponsiveTabs,
    withQueryLoop,
    withTypography,
    withColors
)( GridInspectorControls );
```

tomusborne: I like that idea way more, and each panel still has its own filters we can make use of in GBP.

tomusborne:
Existing Grid blocks using 1.5 switch to the selector even though they have content using this PR: https://www.screencast.com/t/NZvwQ1wk

tomusborne (approved):
Works great as a framework - ready to extend!

PR #167, opened 2021-11-15
Make use of useSelect and useDispatch in Container block controls

tomusborne:
This removes the `wp.data` functions from our Container block controls and makes use of `useSelect` and `useDispatch`.


JeanPaiva (approved):
All good.

PR #168, opened 2021-11-15
Move button url options to toolbar

### LINKED ISSUE #155: Overflow Hidden ( Pseudo Background ) blocks Button URL input field
Screenshot 2021-10-30 at 15 25 12

### Steps to reproduce
1. Add a Container Block and set the Background Image to Pseudo Element.
2. Add a GB Buttons block, make sure the block is near the bottom ( or right edge ) of Container.
3.  Select block and try to access the URL field. 

Question is whether we need to Add Overflow Hidden to Container Block in the editor ?

tomusborne:
Closes #155

This moves the Button block URL options into a toolbar item instead of a floating input that appears automatically.
button-link


JeanPaiva (approved):
Works really well and the code is good. I commented on the Slack channel about a possible feature, we can add it later tho.

PR #169, opened 2021-11-16
Use apiVersion: 2 in Buttons

tomusborne:
This introduces the `apiVersion: 2` method to the Button Container and Button blocks: https://make.wordpress.org/core/2020/11/18/block-api-version-2/

It removes unnecessary elements from the DOM and also allows us to simplify our editor CSS.

This change makes it so the Button URL input isn't focusable, but that's fixed in #168.


JeanPaiva (approved):
All good.

PR #170, opened 2021-11-16
Improve grid column horizontal gap

### LINKED ISSUE #41: UX/UI issue: Gutenberg Toolbox goes beyond the viewport when on smaller screens.
a potential bug.  https://pasteboard.co/JI12rsM.jpg
 
 This occurs on smaller screens when there's a significantly large horizontal gap value assigned to a Grid Block.

tomusborne:
Closes #41 

This adds the `gb-grid-column` wrapper to the Container if it's a grid item. This allows us to use the same horizontal gap method in the editor as we do on the frontend and fixes an issue where the block toolbar was being affected by the gap and not staying attached to the Container itself.

It also switches the Container and Grid blocks to `apiVersion: 2` to simplify the HTML in the editor and remove some CSS (same as we did with Buttons in #169)


src/blocks/container/components/GridItem.js
```diff
@@ -0,0 +1,16 @@
+import classnames from 'classnames';
+
+export default function GridItem( { isGrid, uniqueId, children } ) {
+	if ( ! isGrid ) {
+		return children;
+	}
+
+	return <div className={
+		classnames( {
```
JeanPaiva, line 9: `classnames()` is really necessary here? Both classes will be always present on the div. I think doing directly the classes is more cleaner.

```js
<div className={ `gb-grid-column gb-grid-column${ uniqueId }` }>
```

src/blocks/container/edit.js
```diff
@@ -175,43 +179,48 @@ const ContainerEdit = ( props ) => {
 				googleFontVariants={ googleFontVariants }
 			/>
 
-			<Element
-				tagName={ filterTagName( applyFilters( 'generateblocks.frontend.containerTagName', tagName, attributes ) ) }
-				htmlAttrs={ htmlAttributes }
+			<GridItem
```
JeanPaiva, line 182: Since it has short props you can inline this.

JeanPaiva (approved):
All good.

PR #171, opened 2021-11-16
Use apiVersion: 2 in Headline

tomusborne:
Same as #169 and #170 - this PR converts the Headline block to use `apiVersion: 2`.

Something interesting I noticed when testing is that the `className` attribute is saved automatically when using `useBlockProps()` and `useBlockProps.save()`, so we had to remove the manual addition of that attribute in `htmlAttributes`.


JeanPaiva (approved):
All good.

PR #172, opened 2021-11-16
Keep isGrid and gridId current

tomusborne:
This PR keeps track of the `isGrid` and `gridId` attributes and keeps them current.

This means we can move Container blocks inside and outside of the Grid block and the attributes will update themselves.

This is also useful if the `uniqueId` of the Grid block changes for some reason - it will also update the `gridId` attribute for the Container inside of it.


src/blocks/container/components/InspectorControls.js
```diff
@@ -100,23 +101,38 @@ export default ( props ) => {
...
-		gridContainerId = '';
-
-	if ( typeof getBlockParents === 'function' ) {
-		parentBlockId = getBlockParents( clientId, true )[ 0 ];
+	useEffect( () => {
+		const parentBlockId = getBlockParents( clientId, true )[ 0 ];
 
 		if ( parentBlockId ) {
-			parentBlock = getBlocksByClientId( parentBlockId );
+			const parentBlock = getBlocksByClientId( parentBlockId );
```
JeanPaiva, line 108: Maybe you can create a hook `useParentUniqueId` to extract the logic of retrieving that from here. Check `useInnerBlocksCount` as reference.

tomusborne: Would this be so we can re-use it for other things? In this particular instance we need to know the parent block name and the parent block unique Id.

JeanPaiva: Well, in this case, you can leave as is. :)

JeanPaiva (approved):
All good.

PR #173, opened 2021-11-16
Fix query grid selectors

tomusborne:
Due to the `apiVersion: 2` changes we made we had to add a wrapper around the Query grid items so we can properly apply the horizontal gap CSS.


JeanPaiva (approved):
All good.

PR #174, opened 2021-11-17
Dynamic content framework

JeanPaiva:
Adding dynamic content to Headline block.

**Known issues**
- On selecting different "post preview" the screen is jumping
- ~~Sometimes the Posts select field does not show content, cause it didn't have finished loading the posts.~~ Fixed

**Features to be added**
- More control for the excerpt
- Post date should have format field and choose between post creating or update.


tomusborne:
If we were to do separate blocks, what would we need?:

* Post title
* Post date
* Post meta
* Post comments link
* ?

When it comes to post content/excerpt, what happens if a user uses a heading or a paragraph element type?

A lot of the PHP is already complete over here: https://github.com/tomusborne/generateblocks/tree/implement-dynamic-data-control

That PR has a lot of the necessary options built as well for copy/pasting (post date/updated options etc...)

PR #175, opened 2021-11-29
Create query loop block

JeanPaiva:
This PR is creating the new block for query loop


tomusborne (approved):
Nice and clean - works great.

PR #177, opened 2021-12-01
Make container appender relative

tomusborne:
In WordPress 5.9 (beta 1 now available) they switched the appender to `position: absolute`.

Relative works better in our Container, so this PR switches it back.

Before: https://www.screencast.com/t/WAoUo6kaW
With this PR: https://www.screencast.com/t/EE9LrV5X1JB


JeanPaiva (approved):
All good

PR #178, opened 2021-12-01
Fix color picker UI

tomusborne:
In WordPress 5.9 they switched to react-colorful for the color picker. This resulted in some UI changes that makes our color picker really ugly:
https://www.screencast.com/t/8V9GeENWy
https://www.screencast.com/t/DjR8yGFfmZK

This PR adds some small adjustments to bring it back to normal:
https://www.screencast.com/t/h7G6c0iPz
https://www.screencast.com/t/5leIIf0XhK

We should test to make sure these changes not only fix things in 5.9 but don't break anything in 5.8 as 1.4.1 will be released 1 week before 5.9.

These changes will also need to be applied to GB Pro and GP.


JeanPaiva (approved):
All good.

PR #180, opened 2021-12-09
Add inline background image option

### LINKED ISSUE #58: Feature Request: Add "inline element" style to CSS print method.
Context:
https://wordpress.org/support/topic/inline-background-image/

To be able to lazyload background images. 
https://docs.wp-rocket.me/article/1141-using-lazyload-in-wp-rocket

tomusborne:
Closes #58 

This adds an option to the Container block to add the background image URL to the inline element instead of the compiled CSS.

It does two things:

1. If the background image is set to `Element`, it will output the background image URL as a `background-image` style attribute.
2. If the background image is set to `Pseudo Element`, it will output the background image URL as a CSS variable style attribute, and reference that attribute in the compiled CSS.

This feature is good for users using CDNs, as these URLs will be picked up by those CDN plugins. It will also be necessary for our Query Loop block to reference dynamic background URLs for each post in the loop.

@JeanPaiva feel free to clean anything up/make components if necessary 🙂


tomusborne:
@JeanPaiva not sure what's up with the build failure here. Any ideas?

JeanPaiva:
@tomusborne managed to fix the CI. Had to force the node version as the "ubuntu-latest" image now is installing the latest version of node. Also, had to add two deps to make the tests and lint pass. Also, had to add the .lock file for the npm.

tomusborne:
@JeanPaiva Nice, thank you! Weird how it just became a problem randomly.

JeanPaiva (approved):
All good. Performance enthusiasts will love this feature

PR #182, opened 2021-12-10
Fix undefined option when saving settings

### LINKED ISSUE #181: Bug: undefined option
When switching between embedded and exteral css, the process throws up a warning (in debug mode) about an undefined variable in generateblocks/includes/class-rest.php:134

This only happens if the settings have never previously been saved.

This is the HTTP call that generates the error:
https://mydomain.com/wp-json/generateblocks/v1/settings?_locale=user

My theme uses the "whoops" library, which causes this warning to freak it out. 

Adding a check for the existence of the current setting key fixes it.

```
line 134:
if ( isset($current_settings[ $name ]) && $current_settings[ $name ] === $new_settings[ $name ] ) {
      ....
}
```

Hope that helps. It's a great plugin - it'd be lovely to not have to patch that each time it updates!

cheers,
Chris

tomusborne:
Closes #181

This fixes an undefined index error when saving our options for the first time.

To test, delete the `generateblocks` option from `wp_options` and then click the Save button in the GB settings area.

### DIFF
--- a/includes/class-rest.php
+++ b/includes/class-rest.php
@@ -131,7 +131,7 @@ class GenerateBlocks_Rest extends WP_REST_Controller {
 
 		foreach ( $new_settings as $name => $value ) {
 			// Skip if the option hasn't changed.
-			if ( $current_settings[ $name ] === $new_settings[ $name ] ) {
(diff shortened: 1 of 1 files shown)


JeanPaiva (approved):
All good.

PR #187, opened 2022-01-03
Implement terms query loop

JeanPaiva:
This is adding the query loop for terms.

**Headline is only a workaround to display some content, later we will improve Headline to display more content from terms.**

**Currently working to make child queries a thing**


tomusborne:
Looking good so far! Code looks great.

Working great to display a simple list of terms in editor/frontend.

JeanPaiva:
Putting this on hold for now, we need to find a more solid structure to handle child loops.

PR #190, opened 2022-01-05
Moving bgOptions migration to migration hoc

tomusborne:
Instead of directly changing the attributes on load, this adds the migration using `setAttributes` to the migration hoc.

Typically this shouldn't be an issue anymore, but good to keep for now in case users saved their `bgOptions` before these two options were added.


tomusborne:
I'm guessing the check failure is related to the one we faced here: https://github.com/tomusborne/generateblocks/pull/180

Would be good to get that PR merged so we can update this one with the changes.

JeanPaiva (approved):
All good.

PR #191, opened 2022-01-05
Add withDocumentation hoc

tomusborne:
This moves the Documentation panel from each block into a HOC that filters into the block list.

Originally I used `compose` within `InspectorControls.js` (like the other HOCs), but realized it didn't give us the ability to filter in other controls above this one (https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/#editor-blockedit).

Using `addFilter` here with a high priority allows us to keep the Documentation panel at the bottom of the list of panels.

Perhaps there's a way to ensure it remains at the bottom using `compose`?


JeanPaiva:
I'm getting some errors with old blocks, the old GB page has plenty of this error:

withDoc-error

blockData-undefined

tomusborne:
@JeanPaiva should be good now.

JeanPaiva (approved):
All good.

PR #193, opened 2022-01-06
Fix block margins in editor

tomusborne:
Since we moved to `apiVersion: 2`, our block margin settings have been overwritten by our static CSS removing the core margin from the blocks.

This allows our margin settings to work in the editor while still removing the core margin from the blocks.


tomusborne:
There are still some issues here before we merge. 

Since each block no longer has a wrapper and all blocks are centered with the following CSS, the margin options don't work for root blocks:

```
.editor-styles-wrapper .wp-block {
    margin-left: auto;
    margin-right: auto;
}
```

Not sure if a solution yet.

tomusborne:
This is better, but not perfect.

Since WP adds `max-width` and `auto` margins to `.editor-styles-wrapper .wp-block`, we run into issues when our blocks have no parents (in the root editor).

For example, add a Headline block to the root editor, then toggle the "Inline Width" option.

It seems like some of our blocks could use editor-only wrappers with `useBlockProps`  but only when they're a root block.

tomusborne:
Potential fix here: https://github.com/tomusborne/generateblocks/pull/193/commits/c6eabb8d129a1c31896d0b1b89d9625a152afa73

Basically, we add a conditional wrapper around our blocks (that use `margin`) when they're in the root of the editor. This way we don't add unnecessary divs when they're inside other blocks.

JeanPaiva (approved):
All good. working as expected.

PR #194, opened 2022-01-09
Add missing name props to InspectorControls

tomusborne:
This fixes an issue where the Global Styles toggle stopped showing up for some blocks in the 1.5.0 branch.

This is because we depend on the `name` prop when filtering into the `InspectorControls`: https://github.com/tomusborne/generateblocks-pro/blob/master/src/extend/global/global-styles/index.js#L71

However, the component wasn't passing `name` through it. In some of them we pass `{ ...props }`, so it didn't break.

I wonder if this is an issue in more areas? Perhaps it makes sense passing all of the props through these components? Or is that a performance issue?


JeanPaiva (approved):
All good. You can pass all props down, the reason I didn't was to be clear about what was needed by the component, but there's no performance issue passing all props.

tomusborne:
@JeanPaiva just updated it a bit. Passed `...props` into the `InspectorControls` components and removed what I know is added in props already. This should ensure the filters inside those components continue to get the props they need.

JeanPaiva (approved):
All good

PR #202, opened 2022-01-27
Update wp packages

JeanPaiva:
Updating the packages. Also, I'm now forcing versions, since the packages are updated very frequently by the core team, sometimes patches can break on new `npm install` this way we are always using exact same version. We must keep updating these packages on every major version.


tomusborne (approved):
Good idea, much prefer having to manually change version numbers.

JeanPaiva:
@tomusborne PR is ready. Updated the @wordpress/scripts package and I fixed a bunch of lint issues.

tomusborne:
Looks good to me! `@emotion/react` instead of just `react`?

JeanPaiva:
I think there were errors on build/start cause this package was not found

PR #203, opened 2022-02-01
Extract icon HTML in dynamic headline

tomusborne:
Getting the icon HTML in our dynamic headline isn't super easy.

Our current solution is to convert the attribute to a string:

```
icon: {
    type: 'string',
    default: '',
},
```

This saves the icon HTML into the attributes which we can then use.

The problem with our implementation here is that it always converts that attribute to a string even when we're using a static headline. This results in a ton of unnecessary mess in our saved content: https://www.screencast.com/t/bi0mdwqgu4k

This PR makes it so the icon is still just saved as HTML - not saved in the attributes: https://www.screencast.com/t/VF9EWe8IQr

On the frontend, we extract the icon HTML using `DOMDocument` and output it.


includes/functions.php
```diff
@@ -797,3 +797,59 @@ function generateblocks_is_valid_date( $date, $format = 'Y-m-d\TH:i:s' ) {
...
 	return ( $dateTime && $dateTime->format( $format ) === $date );
 }
+
+/**
+ * Extracts the icon element from our content.
+ * This is useful when using icons in dynamic blocks.
+ *
+ * @param string $content The content to search through.
+ */
+function generateblocks_get_static_icon_html( $content ) {
```
JeanPaiva, line 807: This is a beautifully crafted piece of code. Complex but you made it easy to read!

JeanPaiva (approved):
All good. Really good job on this PR

PR #206, opened 2022-02-03
Fix missing editor styles in Firefox

tomusborne:
This PR fixes an issue where we lose all block styling in the editor using Firefox when in Tablet or Mobile modes.


JeanPaiva (approved):
Working fine on Firefox. `generateblocks-css` and `generateblocks-inline-css` loaded correctly inside the iframe.

PR #216, opened 2022-02-11
Add dynamic content to GB buttons

### LINKED ISSUE #211: Add dynamic content to GB buttons
### Description
We must implement dynamic content for the buttons.

JeanPaiva:
#close #211 

- [x] Once the user selects link type we must disable toolbar link selection
- [x] Don't add LinkWrapper inside the button


tomusborne:
Did 2/3 of to-do list.

Not sure `Once the user selects the link type we should set the link to` has to be done. I think we can just leave the link value empty.

JeanPaiva (commented):
@tomusborne this PR is complete. Ready for review/test.

tomusborne (approved):
Made one more change so the Button link toggle doesn't display if the Button is dynamic but has no link type selected.

PR #229, opened 2022-02-25
Fix button not attributes changing

### LINKED ISSUE #228: Re-usable button blocks need updating on every load
### Description
There's an issue with the Button or Button Container blocks right now when you add them as a Re-usable block.

Something is tricking the editor into thinking the HTML or attributes are being changed on load.

This issue is present in `master` and `1.5.0`.

### Steps to reproduce
1. Add Button block to page
2. Save it as a reusable block
3. Publish the page
4. Refresh the editor, a dot will appear in the Update button saying the reusable block needs to be saved even though nothing has changed.

#### Expected behavior
If nothing in the reusable block changes, it shouldn't prompt the message to update it.

#### Actual behavior
It always displays as needing to be updated even when nothing has changed.

#### References: 
https://wordpress.org/support/topic/reusable-blocks-want-to-save-changes/

tomusborne:
Closes #228

We forgot to define the `blockVersion` attribute in the Button block. This was causing our migration hoc to manually add the attribute on every load, which was making the editor think something had changed and needed to be saved.


JeanPaiva (approved):
Approved. Nice catch!

PR #232, opened 2022-03-01
Improve query appender

tomusborne:
This PR adds a custom Query Loop block appender that adds the pagination if it hasn't already been added.

It also unlocks the template so users can remove the pagination block, but locks the individual Post Template and Container blocks so they can't be removed or moved.

It also adds some frontend classes to the query wrapper and query loop items.


JeanPaiva (approved):
Approved, great job!

PR #233, opened 2022-03-02
Implement pagination numbers

tomusborne:
Blocked by #232 

This PR implements the pagination numbers feature on the front end.

I also moved it from the dynamic link to dynamic content option as I think it makes more sense there (maybe?).


JeanPaiva (approved):
Approved. Added 1 commit, to fix the block label for page numbers and also removed the "add button" when button container is the pagination, users still can drag buttons inside of the pagination tho.

PR #239, opened 2022-03-09
Improve query block UI

### LINKED ISSUE #166: Make responsive tab a HOC for Inspector controls
Today, all blocks have the ResposiveTabs hardcoded inside the inspector controls, the ideal would be possible to manage the position of responsive tabs, to add tabs later/after as we see fit.

Extracting the logic from all blocks to use `withResponsiveTab` HOC we can reduce code duplication and implement a way of handling tabs position easier. Later we might have to do same for all inspector controls to reduce the size of the files.

tomusborne:
closes #166 

This PR improves the Query Loop block UI/UX by doing the following things:

- [x] Makes the Responsive Tabs a HOC so they're always on top. Opted for the `addFilter` method to force it above all other HOCs by setting the priority to `1` (similar to the Documentation HOC).
- [x] Putting Grid block settings inside a Layout panel
- [x] Placing the Query Parameter panel on top of the Layout panel and opening it by default
- [x] Various small tweaks to the parameter list UI
- [x] Making the block icons turn white when selected in the List View (they blended into the blue before)


JeanPaiva (approved):
I liked this improvements, approved.

PR #242, opened 2022-03-11
Add new color component

tomusborne:
This rebuilds our color component from scratch using `react-colorful`.

It also implements a `ColorGroup` control that can be filtered. This allows GB Pro to add to existing color groups without needing its own updated color component: https://github.com/tomusborne/generateblocks-pro/pull/82


src/components/color-group/index.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+import { applyFilters } from '@wordpress/hooks';
+
+/**
+ * Internal dependencies
+ */
+import ColorPicker from '../color-picker';
+import './editor.scss';
+
+export default function ColorGroup( props ) {
```
JeanPaiva, line 12: Awesome component with clean props API. Great work!

JeanPaiva (approved):
Approved! Such nice improvements

PR #244, opened 2022-03-13
Add new number component

tomusborne:
This adds a new number component that includes presets.

We can now define one component for all three devices, which cuts down on the amount of code needed.


src/components/number-control/index.js
```diff
@@ -0,0 +1,178 @@
...
+	const attributeNames = {
+		value: attributeName,
+		unit: attributeName + 'Unit',
+	};
+
+	if ( 'Desktop' !== device ) {
+		attributeNames.value += device;
+	}
+
+	const allPresets = applyFilters(
```
JeanPaiva, line 50: Users will love to configure their presets for fast usage, really powerful!

JeanPaiva (approved):
Approved. Again, great refactor and huge UI improvement!

PR #245, opened 2022-03-14
Keep block-specific panels open

tomusborne:
This makes it so block-specific panels remain open when moving between those blocks.

For example, if you open a Headline block and open the Spacing panel, then switch to a completely different Headline block, the Spacing panel will remain open.

Right now it's using `localStorage` so it will remember your last open panels for each block even if you exit the browser. This means the next time you log in those same panels you left open for each block will still be open.

We could use `sessionStorage` instead if we want to forget which panels are open once the browser is exited.


JeanPaiva (approved):
Approved.

PR #247, opened 2022-03-15
Improve panel memory

tomusborne:
This improves our panel `localStorage` to keep memory of whether panels are open OR closed. This is helpful when `initialOpen` is set to `true` (Container Layout panel) and you close it. This ensures it remains closed even if you toggle away or change device preview.


JeanPaiva (approved):
Approved

PR #248, opened 2022-03-15
Add memory to custom input state

tomusborne:
This adds memory to the custom input state.

For example, if you want to change the `fontSize` value in the Headline block and you input a custom value instead of using the presets, it will remember this as your preference for all `fontSize` attributes and will display the custom input instead of the presets.


JeanPaiva (approved):
All good. I believe some users will be confused with this cause we are deciding to "sync" all headlines for them, maybe a filter to enable/disable this behavior?

PR #249, opened 2022-03-15
Implements GenerateBlocks Image Block

JeanPaiva:
Implements GenerateBlocks Image Block

- [x] Need to check all controls we need
- [x] Fix bug when loading normal image
- [x] Add spacing attributes
- [x] Test within content, query loop, core query loop
- [x] Test and test


JeanPaiva (commented):
The image post meta is not working.

tomusborne:
> The image post meta is not working.

That should be fixed here: https://github.com/tomusborne/generateblocks/pull/290

JeanPaiva (commented):
Approved. This block turned out to be very awesome!

PR #253, opened 2022-03-20
Add dimension options to image block

tomusborne:
This adds sizing options to the image block.

It includes:

1. Image size (full, large etc...)
2. Width
3. Height
4. Object fit

Width, height, and object-fit all have responsive settings. Tablet/mobile width/height depend on object-fit being set at some point as the width/height HTML attributes obviously can't change depending on the device.

One bug I haven't figured out yet is the updating of the width/height depending on the size option when using a dynamic image. Works fine for regular images.


JeanPaiva (approved):
Approved. Pushed a commit to fix JS error when selecting image size that did not existed

PR #255, opened 2022-03-21
Check object values in shouldRebuildCSS

tomusborne:
This improved `shouldRebuildCSS` to check for differences within objects inside arrays (the Effects feature in GB Pro).


JeanPaiva (approved):
approved

PR #263, opened 2022-03-24
Fix shape panel spacing issues

tomusborne:
This fixes some spacing issues in the Shapes panel: https://www.screencast.com/t/liIpHqJh6oN

It forces the shapes to be full width and adds margin under the color control.


JeanPaiva (approved):
Approved

PR #269, opened 2022-03-25
Add color picker groups

tomusborne:
This changes the way our `ColorPicker` and `ColorGroup` components work.

They now group sets of colors so we can display them on the same line. This makes better use of the available horizontal space.

color-groups


JeanPaiva (approved):
Approved

PR #270, opened 2022-03-25
Check for false number values before parsing

tomusborne:
This prevents `NaN` from displaying when the number field blurs with no value.

For example, focus the Button font size input and then click away.

nan

It only happens for the Button right now because the default is `false`.


JeanPaiva (approved):
Approved

Issue #204, opened 2022-02-02
Container full witdh / contained width preview bug in Gutengerg editor


simkoG:
### Description
The Container block has an incorrect preview in the external editor while editing the block/page. The "Full width" is shown as "Contained width" and vice versa. 
 
### Steps to reproduce
1. Install a new WordPress instance (latest version).
2. Install and activate GeneratePress and GenerateBlocks. 
3. Create a new page.
4. Add a "Container" block. 
5. The default setting for the section under Layout > Container is "Full width". 
6. The preview will render the block with "Contained width". 
7. Change the Layout > Container to "Contained width", the preview of the block will be full width within the editor. 

#### Expected behavior
"Full width" preview as full width, 
"Contained width" preview as contained. :)

#### Actual behavior
Opposite of the expected behavior. 
Video: [https://i.imgur.com/kx2YTwg.mp4](https://i.imgur.com/kx2YTwg.mp4)

JeanPaiva:
Hey, @simkoG thanks for your input. I added this to the backlog for our next release.

tomusborne:
Hi @simkoG - I believe this has been fixed in our 1.5.0 branch: https://github.com/tomusborne/generateblocks/tree/release/1.5.0

Let me know if you're able to take a look or not.

Thanks!

PR #272, opened 2022-03-27
Fix outer container width in editor

tomusborne:
While testing #204 I noticed that the container width option wasn't applying to the container when the container is set to Contained width.

This is due to our switch to `apiVersion: 2`. We can now apply the width directly to the `gb-container` element.


JeanPaiva (approved):
Approved

PR #273, opened 2022-03-28
Improve shouldRebuildCSS

tomusborne:
The current implementation wasn't working when an array had multiple objects.

For example, go to "Effects" and add more than one effect to a specific type. `shouldRebuildCSS` wasn't triggering for any changes beyond the first added effect.

I think this should handle all effects as well as simple arrays/objects/values.


JeanPaiva (approved):
Approved. A cleaner way would be to use `Lodash.isEqual`

https://lodash.com/docs/4.17.15#isEqual

You can add the function or the full package: `npm install lodash` or `npm install lodash.isequal`

PR #276, opened 2022-03-29
Add new block icons

tomusborne:
This adds new block icons for our new blocks (Query Loop and Post Template) and updates the other block icons to match their style.


JeanPaiva (approved):
Approved. Beautifully crafted icons, really liked

PR #278, opened 2022-03-30
Add inherit query from template option

### LINKED ISSUE #261: Query loop: Inherit from template
We need a "Inherit query from template" feature for our Query Loop block.

tomusborne:
Closes #261 

This adds an option to inherit the query from the current template. This is necessary for FSE where we're adding a query for a template, or in GP Premium when we add a "Loop Template" Block Element.

I'm not sure if there's anything we need to do in the editor here. Might be worth checking what core does.


tomusborne:
It will also be useful with the Loop Template block element: https://github.com/tomusborne/gp-premium/pull/320

If you create a new Loop Template and apply it to a specific category using Display Rules it should all work.

JeanPaiva (approved):

PR #284, opened 2022-04-04
Use Grid block as Post Template

### LINKED ISSUE #259: Move grid to post template
For a better design we should move all grid options to post-template.

tomusborne:
Closes #259 

This removes the Post Template block and replaces it with the Grid block.

Doing this has a couple of benefits:

1. Removal of the Post Template block. Its only purpose was to act as a container for the loop items.
2. Pagination now sits outside of the Grid block so we don't have to worry about layout issues.


JeanPaiva (approved):
Approved

PR #285, opened 2022-04-05
Add Query Loop templates

### LINKED ISSUE #214: Query Loop starter templates
### Description
In the same way, we have predefined templates for the Grid block, we need to have some initial templates for the Query loop block.

tomusborne:
Closes #214 
Blocked by #284 

This adds a placeholder when you first add the Query Loop block with a few templates.

query-loop-templates

We can add some more once we have the Image block.


JeanPaiva (approved):
Approved. Needs to add option with pagination.

PR #289, opened 2022-04-11
Move pagination appender to BlockControls

tomusborne:
This moves the Query Loop pagination appender to the Block Toolbar.

add-pagination


JeanPaiva (approved):
Approved.

PR #290, opened 2022-04-12
Add dynamic headline caption

tomusborne:
This replaces our RichText caption with the Headline block.

The Headline block as a caption can be used to:

1. Display static text with a static or dynamic image.
2. Display the dynamic caption with a static or dynamic image.

This also adjusts how we get the image data in the editor. We always just fetch the image ID now and we use `getMedia` where needed.


JeanPaiva (approved):
Approved.

PR #298, opened 2022-04-19
Add debounced state hook for the headline and button richtext component

JeanPaiva:


tomusborne (approved):
Feels much nicer in the Query Loop now.

PR #299, opened 2022-04-19
Add Query Loop Block, Image Block, & Dynamic Data

tomusborne:
This is a huge PR that adds a new Query Loop block and dynamic data to our Headline and Button blocks.

Our Query Loop block can be used with core blocks (Post title, Post Content, etc...), or with our dynamic blocks. Our dynamic blocks can also be used with the core Query Loop block.

It also adds our very own Image block, which can be static or dynamic.

Any testing/feedback is greatly appreciated!


src/blocks/headline/components/BlockControls.js
```diff
@@ -16,38 +17,57 @@ export default ( props ) => {
...
+						setAttributes( { alignmentMobile: value } );
+					} }
+				/>
 			}
 		</BlockControls>
 	);
-};
+}
+
+export default memo( HeadlineBlockControls, ( prevProps, nextProps ) => {
```
tomusborne, line 61: Is `memo` needed here or was it just tested to see if it would make a performance difference?

src/blocks/image/components/AnchorTag.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable jsx-a11y/anchor-is-valid, react/jsx-no-target-blank */
```
tomusborne, line 1: Don't like disabling eslint like this but couldn't figure out a better way.

src/extend/dynamic-content/hoc/withPreviewContent.js
```diff
@@ -0,0 +1,10 @@
+export default function withPreviewContent( WrappedComponent ) {
+	return function( props ) {
+		return (
+			<>
+				<WrappedComponent { ...props } />
+				<div>{ 'previews buttons will be here' }</div>
```
tomusborne, line 6: Unnecessary file for now?

JeanPaiva: Removed this file and `withContentLink`.

package.json
```diff
@@ -20,22 +20,25 @@
 	"dependencies": {
 		"@emotion/react": "^11.7.1",
 		"classnames": "^2.3.1",
-		"dompurify": "^2.2.8"
+		"dompurify": "^2.2.8",
+		"use-debounce": "^7.0.1"
 	},
 	"devDependencies": {
```
JeanPaiva, line 26: I think its good time to update all packages.

src/blocks/image/components/AnchorTag.js
JeanPaiva: I think its the only way for this specific case.

PR #301, opened 2022-04-20
Query code cleanup

tomusborne:
Some more tweaks after another review of the code diff.


src/components/icon-picker/index.js
```diff
@@ -86,6 +86,10 @@ class IconPicker extends Component {
 								setAttributes( {
 									'hasIcon': true, // eslint-disable-line quote-props
 								} );
+							} else {
```
tomusborne, line 89: This was removed by mistake in #300

src/components/icon-picker/index.js
```diff
@@ -86,6 +86,10 @@ class IconPicker extends Component {
...
+								setAttributes( {
+									'hasIcon': false, // eslint-disable-line quote-props
```
tomusborne, line 91: This was removed by mistake in #300

PR #308, opened 2022-04-20
Fix query grid

### LINKED ISSUE #305: Query block - Grid not working as expected
There is an issue with the query loop grid when setting container widths for tablet. Setting 50%-50% on desktop doesn't transfer the value for tablet. Also, mobile 100% is not set by default like the normal grid. Works fine on the front end, though.

https://www.loom.com/share/20cda72d17c64fbe85ae5547934a411f

tomusborne:
Closes #305 

This adds a mobile width to the query grid items when you add the template with columns.

It also fixes the width of the grid items on tablet/mobile in the editor.

@JeanPaiva - `useDeviceType` isn't setting the correct state when "Sync Responsive Previews" are turned off as Mike reports here: https://github.com/tomusborne/generateblocks/issues/305#issuecomment-1104034799

Couldn't figure out why. If you disable that option and then go into the editor and switch between devices using our buttons, you'll see `deviceType` is always `Desktop`.


JeanPaiva (approved):

JeanPaiva:
@tomusborne this PR is good. The device type bug will be handled by #312

PR #333, opened 2022-04-22
Only set query attributes if needed

tomusborne:
This PR fixes two things:

1. It fixes a bug where if you add two queries to a single page with different parameters, then multi-select both blocks, the parameters from one block copy over to the other.

https://user-images.githubusercontent.com/20714883/164589979-ac56d55d-5338-4140-a044-d856ad61001d.mp4

2. It fixes a bug where if you simply select the Query Loop without making any changes, the browser confirms your exit as it thinks something has changed.

https://user-images.githubusercontent.com/20714883/164590203-f3ce4ee9-3ff0-4104-b9e9-cd1d23412b4e.mp4


JeanPaiva (approved):
The browser confirm is not entirely fixed, if you select the second post it will trigger the confirm. But that's cause the "Grid" is selected. Probably there's other `setAttributes` being triggered when selecting the grid within query loop. But this PR is approved.

PR #341, opened 2022-04-22
Image fixes and minor features

### LINKED ISSUE #323: Image Block - Caption Dynamic Data Type - only Caption
### Description
There is only Dynamic Data Type of Caption - No Description or other meta ?

### LINKED ISSUE #326: Image block missing current featured image
The image block when set to display the featured image from the current post only works if that featured image has been saved.

https://user-images.githubusercontent.com/20714883/164507540-17c6b2d8-9e9e-4572-91d5-7fb6a027be9e.mp4

JeanPaiva:
This PR does couple of things:

- [x] It adds more image data types (Title, description, alt text). closes #323 
- [x] Fixes the featured image not in sync before save. closes #326 
- [x] Fixes some JS console errors while the image was not fully loaded.
- [x] When no post meta the image is broken.
- [x] Remove "attachment" post type from the selector.
- [x] Remove "nav_menu" from taxonomy selector (query loop).


src/extend/dynamic-content/components/SelectPostType.js
```diff
@@ -8,7 +8,7 @@ export default ( { postType, onChange, value, help } ) => {
 
 	const postTypeOptions = useMemo( () => {
 		return postTypes
-			.filter( ( type ) => ( type.viewable ) )
+			.filter( ( type ) => ( type.viewable && 'attachment' !== type.slug ) )
```
tomusborne, line 11: This breaks what I did here: https://github.com/tomusborne/generateblocks/blob/query-loop-block/src/extend/dynamic-content/InspectorControls.js#L70

The reason I targeted `attachment` above was mainly for editor preview reasons. It pulls the dynamic caption from the `mediaId`. Not sure if there's a better way to do it?

tomusborne: You can replicate the issue like this:

1. Add a static image
2. Add a caption, and enable dynamic data on it

https://user-images.githubusercontent.com/20714883/165123978-d8b7206a-80da-430e-92ba-ffbd9655c7c4.mp4

I suppose we could look at disabling the dynamic data panel for the caption if the image itself isn't dynamic. It's kind of a cool feature that you can use update captions for images and they'll change dynamically even with static images though.

JeanPaiva: @tomusborne my last commit fixes this issue for both static and dynamic images.

JeanPaiva: @tomusborne my last commit fixes this issue for both static and dynamic images.

src/extend/dynamic-content/inspector-controls/ContentTypeControl.js
```diff
@@ -78,6 +78,9 @@ const getOptions = ( name, isCaption ) => {
 				label: __( 'Image', 'generateblocks' ),
 				options: [
 					{ value: 'caption', label: __( 'Caption', 'generateblocks' ) },
+					{ value: 'post-title', label: __( 'Title', 'generateblocks' ) },
+					{ value: 'alt-text', label: __( 'Alt text', 'generateblocks' ) },
+					{ value: 'description', label: __( 'Description', 'generateblocks' ) },
```
tomusborne, line 83: These options are working great in the editor but we need to enable them on the frontend as well.

src/blocks/image/edit.js
```diff
@@ -106,6 +106,7 @@ function ImageEdit( props ) {
 
 			<ContentRenderer
 				{ ...props }
+				featuredImage={ featuredImage }
```
tomusborne, line 109: This fixes the issue if you use the Featured Image panel to add an image. However, the Upload/Media Library options in our image placeholder aren't setting the featured image (`setFeaturedImage`). I wonder if it's worth having that shortcut to set a featured image or if we should only show the Image select placeholder if it's a static image?

JeanPaiva: None of our dynamic blocks set any content, I think we should keep this behavior for the Image as well.

tomusborne: Agreed. We can always add it later if requested. We can handle removing that in another PR.

src/extend/dynamic-content/inspector-controls/ContentTypeControl.js
JeanPaiva: No way, completely forgot about doing the front-end. Its done now.

includes/class-dynamic-content.php
```diff
@@ -569,11 +575,13 @@ public static function get_source_id( $attributes ) {
 			$id = absint( $attributes['postId'] );
 		}
 
+		$image_content_types = array( 'caption', 'post-title', 'alt-text', 'description' );
```
tomusborne, line 578: Could having `post-title` here ever cause a conflict with other dynamic data returning true here? Maybe not because we also check for `dynamicImage` below?

includes/class-dynamic-content.php
```diff
@@ -943,6 +951,40 @@ public static function get_image_caption( $attributes, $block ) {
...
+	public static function get_image_description( $attributes, $block ) {
+		$id = self::get_source_id( $attributes );
+
+		if ( ! $id ) {
+			return '';
+		}
+
+		$media = get_post( $id );
+
+		return $media->post_content;
```
tomusborne, line 985: Think this needs an `isset`? May be playing it too safe lol

tomusborne (approved):

includes/class-dynamic-content.php
JeanPaiva: Yes, we also check for `attachment` post type when using the `postId` attribute

PR #345, opened 2022-04-25
Change isDynamicContent attribute name

tomusborne:
This is a breaking change.

This PR changes the `isDynamicContent` attribute name to `useDynamicData`.

This is the first of a few different attribute changes, but I think it's good to do them separately so it's easier to review each one.


JeanPaiva (approved):
Approved. I created the PR for GBP

PR #348, opened 2022-04-26
Implement focus state for the button block

JeanPaiva:
This adds the focus state for the colors of Button block.

button-block-focus-colors


includes/defaults.php
```diff
@@ -188,7 +188,10 @@ function generateblocks_get_block_defaults() {
 		'textColor' => false,
 		'backgroundColorHover' => false,
 		'backgroundColorHoverOpacity' => 1,
+		'backgroundColorFocus' => false,
+		'backgroundColorFocusOpacity' => 1,
```
tomusborne, line 192: No need for an opacity attribute in new features as we deprecated those in 1.4.

includes/generate-css.php
```diff
@@ -1017,11 +1017,16 @@ function generateblocks_get_dynamic_css( $content = '' ) {
...
 				}
 
-				$css->set_selector( '.gb-button-wrapper ' . $selector . ':hover,.gb-button-wrapper ' . $selector . ':active,.gb-button-wrapper ' . $selector . ':focus' );
+				$css->set_selector( ".gb-button-wrapper {$selector}:hover" );
 				$css->add_property( 'background-color', generateblocks_hex2rgba( $settings['backgroundColorHover'], $settings['backgroundColorHoverOpacity'] ) );
 				$css->add_property( 'color', $settings['textColorHover'] );
 				$css->add_property( 'border-color', generateblocks_hex2rgba( $settings['borderColorHover'], $settings['borderColorHoverOpacity'] ) );
 
+				$css->set_selector( ".gb-button-wrapper {$selector}:active, .gb-button-wrapper {$selector}:focus, .gb-button-wrapper {$selector}.current" );
+				$css->add_property( 'background-color', generateblocks_hex2rgba( $settings['backgroundColorFocus'], $settings['backgroundColorFocusOpacity'] ) );
```
tomusborne, line 1026: Can just output `$settings['backgroundColorFocus']` here.

src/blocks/button/css/main.js
```diff
@@ -126,14 +131,21 @@ export default class MainCSS extends Component {
...
-		.block-editor-block-list__block ` + selector + `:active` ] = [ {
+		cssObj[ `.block-editor-block-list__block ${ selector }:hover` ] = [ {
 			'background-color': hexToRGBA( backgroundColorHover, backgroundColorHoverOpacity ),
 			'color': textColorHover, // eslint-disable-line quote-props
 			'border-color': hexToRGBA( borderColorHover, borderColorHoverOpacity ),
 		} ];
 
+		cssObj[
+			`.block-editor-block-list__block ${ selector }:focus,` +
+			`.block-editor-block-list__block ${ selector }:active`
```
tomusborne, line 142: `.current`?

includes/general.php
```diff
@@ -107,14 +107,23 @@ function generateblocks_do_block_editor_assets() {
...
 
+		$css .= sprintf(
+			'.gb-button.button:active, .gb-button.button:focus {
+				background-color: %1$s;
+				color: %2$s;
+			}',
+			$generatepressDefaultStyling['buttonBackgroundFocus'],
+			$generatepressDefaultStyling['buttonTextFocus']
+		);
+
```
tomusborne, line 126: No need for these changes. The purpose here is to grab the button colors from GP if needed and apply them to a `.button` class if the user adds the GP class to their GB buttons. GP doesn't have focus colors for the buttons, so we can omit this.

tomusborne (commented):
Made some notes.

1. Can get rid all of `Opacity` attributes when dealing with this focus feature as it's new and that opacity slider won't show up.
2. We need to add the attributes to `defaults.php` to prevent undefined notices on the frontend.
3. Instead of `.current` maybe we should do `.gb-button__current` to prevent conflicts.

I agree that the UI will get pretty busy if we add `focus` and `current` options. Merging them like this does make sense. @diggeddy what do you think of this implementation?

diggeddy:
Should we be applying `:focus` here ?
In GP we do this:

`a:hover, a:focus, a:active {
        color: var(--contrast);
 }`
 
And we even go out of our way to ensure `current` and `:focus` are not the same thing in the navigation:

`.main-navigation .main-nav ul li:not([class * ="current-menu-"]):hover > a, .main-navigation .main-nav ul li:not([class * ="current-menu-"]):focus > a, .main-navigation .main-nav ul li.sfHover:not([class * ="current-menu-"]) > a, .main-navigation .menu-bar-item:hover > a, .main-navigation .menu-bar-item.sfHover > a {
        color: var(--accent);
 }`
 
 So why now would we want `current` and `:focus` to be the same thing?

src/blocks/button/css/main.js
JeanPaiva: On the editor we don't have the actual `.current` class applied to our button, so I omitted here.

JeanPaiva:
@diggeddy you think its better to just call it `current` and enable only when the button is inside pagination wrapper? For the editor we still need to target focus for the user preview, cause we don't have the class applied on the editor.

The initial idea was to create controls for current, but it would require `current background/hover` `current text/hover` and `current border/hover` which is 3 more controls.

tomusborne:
Definitely an interesting discussion...

How badly do we need a `focus` option vs using the current `hover` options as the focus (which I think we're doing now)?

If we (for now at least) concede that `focus` is handled by `hover`, we can conditionally add a `current` option and the UI won't get overloaded.

diggeddy:
I don't feel we need `:focus` at this time, and I don't feel its appropriate we use it for the Paging Navigation current color. For now:

> we can conditionally add a current option and the UI won't get overloaded.

tomusborne:
I explored that method here: https://github.com/tomusborne/generateblocks/pull/360

It's only available for the `pagination-numbers` dynamic content type for now.

PR #349, opened 2022-04-26
Improve post excerpt preview

### LINKED ISSUE #339: Headline - Dynamic Excerpt - Editor trims length to GP Blog Settings.
### Description

The Headline Dynamic excerpt is trimmed to GP Blog Excerpt length in the editor.
If you set the GP Excerpt to 55 - then GB excerpt will never display greater length.

It is correct on the front end.

tomusborne:
Closes #339 

This PR was a bit of a challenge. As @diggeddy found in #339 our excerpt preview was taking the excerpt from the theme and using it as a base for our previews. This becomes an issue if we want to increase the `excerptLength` to a larger number than the theme excerpt.

So, this PR does the following:

1. If we're using the more tag or a manual excerpt, use the excerpt.
2. If not, use the content and create our own excerpt from it using our excerpt length.


JeanPaiva:
@tomusborne this is tricky, creating our own excerpt on the fly, its very risky in my opinion and could become a performance issue on big loops. So many unknown variables coming from the content.

The first main issue is blocks that are not allowed for excerpt (eg. our own Headline as excerpt) we would need to remove all this guys. The duplicated content comes from our own headline excerpt on the post.

image

This block is using the excerpt that comes from the rest API and rendering it, its not his job to generate a excerpt on the fly in my opinion. If the theme set that the excerpt is 55, its not a block that will tell otherwise.

JeanPaiva:
@tomusborne this is tricky, creating our own excerpt on the fly, its very risky in my opinion and could become a performance issue on big loops. So many unknown variables coming from the content.

The first main issue is blocks that are not allowed for excerpt (eg. our own Headline as excerpt) we would need to remove all this guys. The duplicated content comes from our own headline excerpt on the post.

image

This block is using the excerpt that comes from the rest API and rendering it, its not his job to generate a excerpt on the fly in my opinion. If the theme set that the excerpt is 55, its not a block that will tell otherwise.

src/extend/dynamic-content/utils/getContent.js
```diff
@@ -86,38 +86,64 @@ function getPostExcerpt( record, attributes ) {
...
+		raw: rawContent,
+		rendered: renderedContent,
+	} = record?.content;
+
+	if ( isProtected || ( ! rawContent && ! renderedContent ) ) {
 		return __( 'No post excerpt.', 'generateblocks' );
 	}
 
-	const renderedText = ( text ) => new window.DOMParser().parseFromString( text, 'text/html' );
+	const buildExcerptFromContent = ( text ) => text
```
JeanPaiva, line 98: Need to remove all blocks that are not allowed for excerpt here too.

tomusborne:
Let's leave this as it is. This PR will likely end up causing more headaches than it solves.

PR #351, opened 2022-04-26
Add image alignment options

### LINKED ISSUE #340: Image Block - Dynamic Image missing align classes on front end.
### Description
Editor alignment options applies an `align*` class to the element.
This applies in the editor but not on the frontend IF the image is set to Dynamic Data Type eg. Featured image.

### Steps to reproduce
1.  Add an Image Block, set it to Dynamic > Post Type > Post --> some post --> Featured Image.2. 
2. Put some content below just to see the floating applied.
3. Set the image block to Align left3. 
4. Check the front end

This probably needs to part of a wider discussion regarding the core Toolbar options.

tomusborne:
Closes #340 

This PR adds alignment options to the image block.

We add:

1. wide/full alignment if the theme supports it
2. left/center/right `text-align` options on all three devices
3. left/right/none `float` options on all three devices (`none` only exists on tablet/mobile)


src/blocks/image/components/BlockControls.js
```diff
@@ -26,6 +27,8 @@ export default function ImageBlockControls( props ) {
 
 	return (
 		<BlockControls>
+			<AlignmentControls { ...props } />
```
JeanPaiva, line 30: This component does need all props from `ImageBlockControls`? On the component itself you use only 3, this can cause unnecessary renders.

includes/generate-css.php
```diff
@@ -1666,6 +1666,17 @@ function generateblocks_get_dynamic_css( $content = '' ) {
 				$css->add_property( 'padding', array( $settings['paddingTop'], $settings['paddingRight'], $settings['paddingBottom'], $settings['paddingLeft'] ), $settings['paddingUnit'] );
 				$css->add_property( 'margin', array( $settings['marginTop'], $settings['marginRight'], $settings['marginBottom'], $settings['marginLeft'] ), $settings['marginUnit'] );
 
+				// Set a flag we'll update later if we disable floats.
+				$disable_float = false;
```
JeanPaiva, line 1670: Filter?

JeanPaiva (approved):

includes/generate-css.php
tomusborne: That flag is mainly there so we don't set `float: none` on tablet **and** mobile. So if we set the flag on tablet we don't set `float: none` again on mobile. We could likely remove it but it would require another condition on mobile to see if the condition to remove the float on tablet was met.

PR #360, opened 2022-04-27
Add current button colors

### LINKED ISSUE #313: Query loop pagination does not have button active state
### Description
When using the pagination for query loop, we need to differentiate the active page.

I think we can add a new color option for "current page" styles of the button.

### Steps for testing
1. Add a query loop block
2. Add navigation to it.
3. Goes to the front-end and click on page two for example.

tomusborne:
Closes #313 
Alternative to #348 

This method adds a "Current" button color option to the button block if we're working with the `pagination-numbers` dynamic content type. We'll want to add more conditions to this in the future (accordions, tabs, etc...).

I also removed the `:visited` selector from the frontend, as I don't believe it's necessary.


src/blocks/button/components/ConditionalColors.js
```diff
@@ -0,0 +1,68 @@
...
+	const {
+		useDynamicData,
+		dynamicContentType,
+	} = attributes;
+
+	if ( 'generateblocks/button' !== name ) {
+		return items;
+	}
+
+	const addCurrentColors = useDynamicData && 'pagination-numbers' === dynamicContentType;
```
tomusborne, line 19: We could add a filter here or rely on the same method we're using to add these colors if we don't have access to this variable.

JeanPaiva: Don't think its necessary, cause its very specific use case.

JeanPaiva (approved):

PR #364, opened 2022-04-28
Remove $block variable from attribute functions

tomusborne:
`generateblocks_attr()` only accepts three parameters. `$block` was added when we were adding them to the available args in the filters, but they're useless in these functions.

An alternative to removing them is to add another arg to the function itself so it accepts them. Might be worth it.


tomusborne:
It would be nice to have access to `$block` in the filter. Going to close this.

PR #365, opened 2022-04-28
Add filter to fallback dynamic images

JeanPaiva:
Filter usage:

Editor filter:
```js
import { addFilter } from '@wordpress/hooks';

addFilter(
	'generateblocks.editor.dynamicImageFallback',
	'generateblocks-pro/dynamic-content/post-record',
	( image, context ) => {
		if ( ! image ) {
			return 'http://generate.local/wp-content/uploads/2008/06/dsc20051220_173257_119-1.jpg';
		}

		return image;
	}
);
```

Front-end filter:
```php
add_filter( 'generateblocks_dynamic_image_fallback', function( $url, $attributes, $block ) {
	return 'http://generate.local/wp-content/uploads/2008/06/dsc20051220_173257_119-1.jpg';
}, 3, 10 );
```


includes/class-dynamic-content.php
```diff
@@ -508,7 +508,7 @@ public static function get_dynamic_image( $attributes, $block ) {
 		$id = self::get_dynamic_image_id( $attributes );
 
 		if ( ! $id ) {
-			return;
+			$id = apply_filters( 'generateblocks_dynamic_image_fallback', '', $attributes, $block );
```
tomusborne, line 511: Should we return after this if it's still empty?

includes/class-dynamic-content.php
```diff
@@ -520,13 +520,9 @@ public static function get_dynamic_image( $attributes, $block ) {
...
 		}
 
-		$html_attributes = generateblocks_parse_attr(
-			'image',
-			array(
-				'id' => isset( $attributes['anchor'] ) ? $attributes['anchor'] : null,
-				'class' => implode( ' ', $classes ),
-			),
-			$attributes
+		$html_attributes = array(
```
tomusborne, line 523: I adjusted this so we don't parse the attributes twice.

src/blocks/image/components/ImageContentRenderer.js
```diff
@@ -98,7 +99,14 @@ export default function ImageContentRenderer( props ) {
 	const parentBlock = getBlockRootClientId( clientId );
 	const currentImage = getDynamicImage( props );
 	const dynamicImageUrl = getMediaUrl( currentImage, sizeSlug );
-	const imageUrl = useDynamicData && dynamicContentType ? dynamicImageUrl : attributes.mediaUrl;
+
+	const dynamicImageFallback = applyFilters(
+		'generateblocks.editor.dynamicImageFallback',
+		dynamicImageUrl,
+		context
```
tomusborne, line 106: Is passing `props` here a bad idea so it's more in line with the frontend filter where we have access to attributes and context/block name etc...?

PR #368, opened 2022-04-28
Fix failing build check

tomusborne:
This fixes the failing build checks in `query-loop-block`.


src/extend/dynamic-content/utils/getContent.js
```diff
@@ -154,18 +154,21 @@ function getPostDate( record, attributes ) {
...
  */
 const getMetaValue = ( metaField, metaValues, emptyNotFoundMessage = false ) => {
 	if ( metaValues && metaValues[ metaField ] ) {
 		const value = metaValues[ metaField ];
+		const notSupportedMessage = ! emptyNotFoundMessage
+			? __( 'Meta value not supported.', 'generateblocks' )
+			: undefined;
 
 		return ( _.isString( value ) || _.isNumber( value ) )
 			? _.toString( value )
```
tomusborne, line 170: Does our post meta value need to be converted to a string here?

JeanPaiva: It does, cause the Rich Text component can't render numbers (eg. ACF range)

JeanPaiva (approved):

PR #369, opened 2022-04-29
Improve usePostRecord hook

JeanPaiva:
This PR improves the `usePostRecord` for dynamic content.

1. It does add a `Loading...` message while there's any request happening.
2. It removes the high initial load by making only necessary requests based on the `content-type`

It does improve performance a bit, cause we load author, comments, terms only when the related content types are selected.
On a query loop with dynamic content only from the post record, we don't do any extra API calls, which is pretty good.


tomusborne:
Great PR - cleans it up a bit.

One thing I noticed here is the editor confirming refresh again without anything changing.

1. Add a dynamic headline, save
2. Refresh, then click on the headline
3. Refresh again to see the confirm message even though no attributes changed

tomusborne (approved):

PR #382, opened 2022-05-02
Show image as loading when uploading

tomusborne:
This PR fixes an issue that happens when you save a page while an image is still uploading.

https://user-images.githubusercontent.com/20714883/166328001-6c1b4cef-0f20-4b91-ae67-0f549f2fda4f.mp4

With this PR it shows the image as loading so the user will wait until it's done uploading.

https://user-images.githubusercontent.com/20714883/166328247-ffb06588-7dbc-4674-9dfe-0872e63b8528.mp4


JeanPaiva (approved):
all good.

PR #403, opened 2022-05-04
Fix excerpt not showing with multi query loops in the same page

### LINKED ISSUE #399: Excerpt not showing with multi query loops in the same page
### Description
The post excerpt is not correctly showing when you have two or more query loops on the page.

two-queries

### Steps to reproduce
1. Add 2 query loops with the post excerpt.

#### Expected behavior
1. Both loops display the excerpt correctly.

#### Actual behavior
1. The second excerpt is not showing.

JeanPaiva:
closes #399


tomusborne (approved):
Nice fix! Double checked for infinite loops and still all good.

PR #410, opened 2022-05-04
Add featured image template

### LINKED ISSUE #406: offer a query loop with featured image as one of the defaults
### Description
It could be not clear for beginer/intermediate users how to add a featured image into the query loop

### Steps for testing
1. Create a default query loop with featured image

#### References:
...

-----

JeanPaiva:
closes #406


tomusborne (approved):

quantumleap33:
@JeanPaiva I think the 20px padding included in the template is not needed. Better keep it as default as possible and maybe increase the grid horizontal gap to 40px.

PR #432, opened 2022-05-06
Add multi select component

### LINKED ISSUE #412: Query Loop Block: Multiple Post Statuses
Please let us select multiple statuses in the post status parameter to show past and upcoming post for example.

JeanPaiva:
closes #412


tomusborne (approved):
Work nice. Tested by querying Draft and Private posts. Works in editor and the frontend.

PR #433, opened 2022-05-06
Allow meta values be created if not enabled on the API

### LINKED ISSUE #413: Metabox custom fields not showing as options when Post meta selected
I am running into an issue with the dynamic data in connection with Metabox custom fields. With the "old" GP dynamic data system for elements, I can select the dynamic icon on the block toolbar, select Post meta, and then enter a meta field name from one of the Metabox custom fields. But now with this dynamic section in the block settings on the right side, I can select Post meta, but the drop-down selection for the Post meta field is not finding the Metabox custom fields, and I can't paste in a custom meta field name. So basically, it seems that I am unable to access Metabox fields with the new Dynamic Data tool section.

Screen Shot 2022-05-04 at 11 00 32 PM

Screen Shot 2022-05-04 at 11 01 44 PM

JeanPaiva:
closes #413


src/extend/dynamic-content/inspector-controls/PostMetaControl.js
```diff
@@ -26,10 +30,12 @@ export default function PostMetaControl( props ) {
 				<AdvancedSelect
 					id={ 'gblocks-select-post-meta-control' }
 					label={ __( 'Post meta field', 'generateblocks' ) }
-					placeholder={ __( 'Post meta field', 'generateblocks' ) }
+					help={ __( 'Live preview is only available to meta exposed to the REST API.', 'generateblocks' ) }
+					placeholder={ __( 'Choose or create meta field', 'generateblocks' ) }
```
tomusborne, line 34: I'm not seeing this placeholder show up. Instead, we could add this to the beginning of `help` above.

tomusborne (commented):
Should we make it so the user can remove a value once added? I was trying to see the placeholder and couldn't remove the choice I already selected.

tomusborne (approved):
This is awesome - great work!

PR #434, opened 2022-05-06
Fix Container vertical alignment in Query Loop

tomusborne:
This fixes an issue reported in our Facebook group: https://www.facebook.com/groups/generatepress/posts/5012822445439959/?comment_id=5012834982105372

The selectors were over-complicated. Simplifying like this fixes the issue and works inside normal grid and Query Loop grids.

To test:

1. Add a Grid with two columns
2. Give the Container a min-height of 500px
3. Add text inside the Container, and set the Container vertical alignment

This works in normal Grid blocks, but doesn't work when using the Query Loop block

This PR should fix the Query Loop block while still allowing the normal Grid block to work as usual.


JeanPaiva (approved):
Cool fix!

PR #437, opened 2022-05-07
Fix loss of device image sizes

### LINKED ISSUE #436: Image Block: Responsive image sizing doesn't work in WP 5.9.
Responsive image sizing doesn't save. See video:
https://www.screencast.com/t/ZPBUvIv4kQ

The issue only exists in WP5.9 as Tom said it works in WP6.0

tomusborne:
Closes #436 

This fixes an issue where the device image sizes were being cleared in WP 5.9. I wasn't able to reproduce in 6.0.

Basically, we prevent our `setAttributes()` hook from running on first render.


leohsiang:
The width/height should carry over when the image is switched:
https://www.screencast.com/t/Cn5jlioG

JeanPaiva (approved):

PR #443, opened 2022-05-09
Improve BlockPreview performance

tomusborne:
This PR improves the performance of our Query Loop by debouncing changes in the BlockPreviews.


src/blocks/query-loop/components/LoopRenderer.js
```diff
@@ -64,8 +72,13 @@ export default function LoopRenderer( props ) {
 		return select( 'core/block-editor' )?.getBlocks( clientId );
 	}, [] );
 
+	const [ innerBlockData, setInnerBlockData ] = useState( setIsBlockPreview( innerBlocks ) );
 	const [ activeContextId, setActiveContextId ] = useState();
 
+	useDebouncedEffect( () => {
+		setInnerBlockData( setIsBlockPreview( innerBlocks ) );
+	}, [ JSON.stringify( innerBlocks ) ], 250 );
```
tomusborne, line 80: We may be able to reduce the delay even more without harming performance.

src/blocks/query-loop/components/LoopRenderer.js
```diff
@@ -49,6 +49,14 @@ function setIsBlockPreview( innerBlocks ) {
 	} );
 }
 
+function useDebouncedEffect( effect, deps, delay ) {
```
JeanPaiva, line 52: Instead of creating your own, you could use the package we added `use-debounce`.

`import { useDebounce, useDebouncedCallback } from 'use-debounce';`

JeanPaiva (commented):
This adds some improvement when typing really fast.

With this we need to remove the bounce we added for the headline/button blocks, cause they are doubling the bounce effect, cause bit more delay.

Also it brings some weird behaviors, for example when the author object is loading, we don't display nothing while the request still hanging, check the image.

image

src/blocks/query-loop/components/LoopRenderer.js
JeanPaiva: Something like this should work
```js
	const debounced = useDebouncedCallback( () => {
		setInnerBlockData( setIsBlockPreview( innerBlocks ) );
	}, 1 );

	useEffect( () => {
		debounced();
	}, [ JSON.stringify( innerBlocks ) ] );
```

src/blocks/query-loop/components/LoopRenderer.js
```diff
@@ -64,8 +72,13 @@ export default function LoopRenderer( props ) {
 		return select( 'core/block-editor' )?.getBlocks( clientId );
 	}, [] );
 
+	const [ innerBlockData, setInnerBlockData ] = useState( setIsBlockPreview( innerBlocks ) );
```
JeanPaiva, line 75: I think its better to start with empty array and set the blocks with effect.
```js
useEffect( () => {
	setInnerBlockData( setIsBlockPreview( innerBlocks ) );
}, [] );
```

PR #485, opened 2022-05-24
Keep empty image alt

### LINKED ISSUE #484: Image Block: Lighthouse flags img elements that don't have alt attributes
[Lighthouse expects](https://web.dev/image-alt/) all `` elements to include an `alt` attribute, even for decorative images wihtout `alt` text.

``

The default WordPress image block produces ``, which is evidently also acceptable. 

The GB Image Block is not currently generating the `alt` attribute for images with no alt text. Lighthouse flags this as a failing element for Accessiblity/SEO. 

``

tomusborne:
Closes #484 

This keeps the empty image `alt` attribute in the image HTML.

Not sure if you have a better idea on how to do this @JeanPaiva - this feels a little rough.


src/blocks/query-loop/components/utils.js
```diff
@@ -1,5 +1,10 @@
-export function removeEmpty( obj ) {
-	return Object.fromEntries( Object.entries( obj ).filter( ( [ idx, value ] ) => { // eslint-disable-line no-unused-vars
+export function removeEmpty( obj, isImage = false ) {
```
JeanPaiva, line 1: Instead of having a boolean argument a cleaner way to do it would be to have an array of allowed empty attributes. That way in the future other components might use same approach. But fine for now.

JeanPaiva (approved):

PR #491, opened 2022-05-24
Set default value placeholder

### LINKED ISSUE #447: [GB Query loop] Default posts per page is incorrect when value is removed from the input field on the backend
Query loop defaults to 10 posts per page on the backend (when no Posts per page have been set, eg, number is removed from the input field) even though the **Settings->Reading->Blog pages show at most** has been set to a different number. Previewing the query loop on the front end shows the correct number of posts from the Reading settings.

JeanPaiva:
We are adding default values as placeholder for the query parameters.

closes #447


JeanPaiva:
@tomusborne added the default for the front-end. On the editor 10 is the REST API default.

includes/general.php
```diff
@@ -327,3 +327,18 @@ function generateblocks_set_inline_background_style( $attributes, $settings ) {
 
 	return $attributes;
 }
+
+add_filter( 'generateblocks_query_loop_args', 'generateblocks_set_query_loop_default_query_args', 10, 1 );
```
tomusborne, line 331: We could add this in `class-query-loop.php`. Would mean we can simplify the function name (`public function set_query_loop_defaults`).

Not sure we need the `1` priority either (unless you have a reason?).

tomusborne (approved):

PR #497, opened 2022-05-25
Fix grid layout selector

tomusborne:
This fixes a bug in WP 6.0 where the Grid layout selector is all mangled.

grid

Its already fixed in GB 1.5.0, but we may want to release 1.4.4 for now as it looks pretty bad for our users using 6.0.

Tested on WP 6.0 and WP 5.9.3.


JeanPaiva (approved):
Approved.

Issue #492, opened 2022-05-24
Block error when duplicating two headline blocks simultaneously


quantumleap33:
There is an error when two or more headline blocks are selected and duplicated.
https://www.loom.com/share/886e48910ed241429a31553cfca24d6d

PR #502, opened 2022-05-26
Fix uniqueID error when multi blocks selected

JeanPaiva:
Closes https://github.com/tomusborne/generateblocks/issues/492


src/utils/is-block-multi-selected.js
```diff
@@ -0,0 +1 @@
+export default ( clientId ) => wp.data.select( 'core/block-editor' ).isBlockMultiSelected( clientId );
```
JeanPaiva, line 1: This is need in order for unit tests mock the return value of the function.

tomusborne: Does it need to use `wp.data` vs importing?

tomusborne (approved):

PR #513, opened 2022-05-30
Fix: Contained container not applying auto margins

### LINKED ISSUE #510: Backend alignment issue
### Description
There is a bug when two containers and stacked vertically on a full-width page.

### Steps to reproduce
1. Create a new page and set the page container to `Full Width`.
2. Insert a container, put some padding all around and choose a color for greater visibility during testing.
3. Insert another container below it and set it to `Contained width`. Put some padding all around and choose a color for greater visibility during testing.
4. Try adding some margins to the last inserted container.

#### Expected behavior
The container should stay at the center of the page.

#### Actual behavior
The container floats to the left. Any margin value (even if `0`) to the second container will cause that to happen. Frontend works fine.

JeanPaiva:
close #510


tomusborne:
This works but I'm unable to see the difference because of the following CSS we have.

container-width

I have no memory as to why I added that. Doesn't really seem necessary - think we can remove.

It sucks that auto-margin is "assumed". That's going to take some undoing/back-compat magic once we allow for better auto margin control.

tomusborne (approved):

PR #544, opened 2022-06-07
GitHub actions

JeanPaiva:
- Fix the JS CI to use cache of node_modules instead of installing dependencies on every action
- Adds manual action to build the GenerateBlocks package
- Adds the action to build dist files on release branches

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -1,0 +1,40 @@
+name: Build package
+
+on:
+  workflow_dispatch:
+    inputs:
+      name:
+        description: 'Package name'
+        default: 'generateblocks'
+        required: true
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-package.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: npm install
+
+      - name: NPM build
+        run: npm run build
+
+      - name: Build zip file
+        uses: actions/upload-artifact@v3.1.0
+        with:
+          name: ${{ github.event.inputs.name }}
+          path: |
```
tomusborne, line 33: Is it possible to package all files except for the ones defined in `.distignore`?: https://github.com/tomusborne/generateblocks/blob/master/.distignore

PR #549, opened 2022-06-08
Fix: Color picker input field

### LINKED ISSUE #540: Jumping cursor in RGBA color pallete value editing
When editing an RGBA color palette value, the cursor jumps at the end of the string. Basically, this happens when inserting or deleting a character. The cursor keeps jumping at the end of the string and I need to use the keyboard left arrow to put the cursor back in place. It happens both in the GP Customizer color palette and the GB one.

https://www.loom.com/share/72bb4f52315843a9a0fd45ff516f66ef

JeanPaiva:
close #540

### DIFF
--- a/src/components/color-picker/index.js
+++ b/src/components/color-picker/index.js
@@ -34,30 +34,11 @@ export default function ColorPicker( props ) {
 		tooltip,
 	} = props;
 
-	const [ isManualInput, setManualInput ] = useState( false );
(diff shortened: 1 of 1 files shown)


tomusborne:
Closing this for now. We can revisit in 1.6.

JeanPaiva:
@tomusborne should we re-open this now?

PR #558, opened 2022-06-10
Tweak: Change image description option value name

tomusborne:
`description` seemed like a very vague option value name. This will break existing Image block captions using the `description` value, but I figure it's best to do this now before things are live.

Not sure it's necessary - open to discussion.


JeanPaiva (approved):
Good call

PR #585, opened 2022-06-20
Quick fixes

JeanPaiva:
PR with some quick fixes before releasing

- Change Query loop grid label to "Post template"
- Change Taxonomies select label to "Select terms"
- Fix query loop per page default value of 10

query-loop-post-template
query-loop-select-terms

### DIFF
--- a/src/blocks/grid/block.js
+++ b/src/blocks/grid/block.js
@@ -52,4 +52,9 @@ registerBlockType( 'generateblocks/grid', {
 		);
 	},
 	deprecated,
+	__experimentalLabel: ( attrs ) => {
+		return attrs.isQueryLoop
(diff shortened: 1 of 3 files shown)


src/blocks/grid/block.js
```diff
@@ -52,4 +52,9 @@ registerBlockType( 'generateblocks/grid', {
 		);
 	},
 	deprecated,
+	__experimentalLabel: ( attrs ) => {
+		return attrs.isQueryLoop
+			? __( 'Post template', 'generateblocks' )
```
tomusborne, line 57: Let's make this a capital T and the Query Loop above it has caps for both words.

tomusborne (approved):

Issue #553, opened 2022-06-08
When drag a Grid Container out of the Grid, it doesn't turn into a normal Container block


leohsiang:
### Description
When drag a Grid Container out of the Grid, it doesn't turn into a normal Container block.
On backend: The container still has Grid container's options, eg. width(%) , but no matter what width you entered, it keeps 100% in width.
On frontend: The container still under a `gb-grid-column`, even there's no Grid block.

### Steps to reproduce
1.  Add a Grid block with 2 containers.
2.  Drag one container out from the Grid block in the list view.

#### Expected behavior
...

#### Actual behavior
...

#### References: 
...

-----

PR #596, opened 2022-06-23
Fix: Container gridId attribute not updating correctly

### LINKED ISSUE #595: Query Loop Columns bug in editor and front end
### Description
Columns not displaying after a single column query loop block is inserted.

Watch this: https://www.screencast.com/t/VyhF3Dc65N24

JeanPaiva:
close #595


tomusborne (approved):
This fixes the reported issue. Hopefully we can find a better solution in 1.6 to improve performance a bit.

This bug was introduced in #554, and the fix here still fixes the original issue (#553).

PR #607, opened 2022-06-27
Tweak: Improve dynamic CSS generation

### LINKED ISSUE #454: Enqueue styles for blocks added in FSE
### Description
If you are doing some Full Site Editing (FSE) and add GenerateBlocks blocks to your templates (i.e. header, footer, page templates) everything looks perfect in the backend when editing, but, on the frontend there are no styles applied because the styles for these are not enqueued (I believe that's the problem anyway).

### Steps to reproduce
1. Enable TwentyTwentyTwo theme
2. Appearance -> Editor
3. Edit Template Parts (Footer) or a Page Template
4. Save and view page (in another browser)

#### Expected behaviour
Styles for GenerateBlocks should show up on the front end in your footer or page templates

#### Actual behavior
The styles only show up in the backend. :(

NOTE: I mentioned this to @tomusborne previously and just thought I should add it here so that it's not forgotten. :)

tomusborne:
close #454 

This PR does a few things:

1. It adds block-specific PHP classes that hold their defaults, CSS and frontend renders.
2. If one of our blocks has been added to a part of a site that's not in the `post_content` it should generate the necessary CSS inline using `<style>` tags above the block HTML.


includes/class-render-blocks.php
```diff
@@ -59,7 +59,7 @@ public function register_blocks() {
 			'generateblocks/container',
 			array(
 				'title' => esc_html__( 'Container', 'generateblocks' ),
-				'render_callback' => array( $this, 'do_container_block' ),
```
tomusborne, line 62: @JeanPaiva how safe is it to remove these functions from the class? Not sure if we should deprecate them.

JeanPaiva: Yes, not sure any user would extend this class.

JeanPaiva (approved):
Seems everything is working fine.

The widget are requires the commit https://github.com/tomusborne/generateblocks-pro/pull/151/commits/773630f2a3db8db20f538829edf4b33f144527ca in order to work, we can merge after the ACF or cherry-pick the commit.

PR #621, opened 2022-07-07
Fix: Query loop taxonomy query include/exclude children

JeanPaiva:
Fix taxonomies/exclude taxonomies parameter to include children terms by default. Also adds a toggle for users to choose what they want.

### DIFF
--- a/includes/class-query-loop.php
+++ b/includes/class-query-loop.php
@@ -172,7 +172,7 @@ class GenerateBlocks_Query_Loop {
 					'field'    => 'term_id',
 					'terms'    => $tax['terms'],
 					'operator' => $operator,
-					'include_children' => false,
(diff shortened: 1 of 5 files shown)


tomusborne:
I'm seeing an issue in the editor where the children aren't included unless you've selected a block inside the Query Loop and the Query Loop at some point.


https://user-images.githubusercontent.com/20714883/178297102-71fc8d69-38ff-483d-b256-d0601b20f098.mp4

JeanPaiva:
> I'm seeing an issue in the editor where the children aren't included unless you've selected a block inside the Query Loop and the Query Loop at some point.
> include-children.mp4

Related to https://github.com/tomusborne/generateblocks/blob/fix-include-children-tax/src/blocks/query-loop/components/InspectorControls.js#L31

tomusborne (approved):

PR #642, opened 2022-07-25
Feature: Improve select components

JeanPaiva:
WIP


tomusborne:
@JeanPaiva Could we do a "Load More" button at the bottom of the list? I assume if we click the button and no results are found we can hide it?

tomusborne:
And there's no way to catch the error and do something with it?

I think this should be targeted for `release/1.6.0` so we have more time to test/improve it if possible.

tomusborne (approved):

PR #661, opened 2022-08-22
Release: 1.6.0

JeanPaiva:
### Definition of done
- [x] Update version: package.json/package-lock.json
- [x] Update version: plugin/style.css header
- [x] Update version: PHP constant
- [x] Update version: readme.txt stable tag
- [x] Update changelog
- [x] Test for PHP parse errors
- [x] Test PHP 5.6+


includes/functions.php
```diff
@@ -893,3 +893,246 @@ function generateblocks_filter_images( $content, $attributes ) {
...
+		[
+			'main' => '',
+			'desktop' => '',
+			'tablet' => '',
+			'tablet_only' => '',
+			'mobile' => '',
+		]
+	);
+
+	$existing_data['main'] .= generateblocks_get_parsed_css( $new_data['main'] );
```
tomusborne, line 972: @JeanPaiva should we add `isset` checks for the `$new_data['xx']` variables in this function? Right now it's just assuming they exist.

PR #668, opened 2022-08-23
Feature: Add filter for query loop enabled post types

### LINKED ISSUE #631: Introduce query loop post type select filter
### Description
Not sure if this is expected behaviour currently with the query loop block, but when I have a custom post type set to publicly queryable false it is not selectable in the select post type dropdown. 

This sort of makes sense, but I might also want to hide a post type from the front-end so a user can't stumble across pages accidently, but then do something custom with the post type using the query loop. 

So I think the post type should still be selectable as a source so we can decide if want to do something custom with it.

### Steps to reproduce
1. Create a new custom post type
2. Set publicy queryable to false
3. Create a new query loop and select source
4. Post type isn't available

#### Expected behavior
I think if we've set publicy queryable to false to hide from the front-end it should still be selectable in the post type source of the query loop for use cases such as customised listings etc.

JeanPaiva:
closes #631 

By default we only display post types that have `viewable` set to true, with this filter users can enable any other post type.

Filter usage: 

```js
wp.hooks.addFilter(
	'generateblocks.editor.query-loop.enabled-post-types',
	'testing-not-viewable-post-types',
	function( enabledPostTypes, postTypes ) {
		const customPostTypes = postTypes.filter( ( postType ) => ( 'custom-slug' === postType.slug ) );

		return enabledPostTypes.concat( customPostTypes );
	}
);
```

### DIFF
--- a/src/extend/dynamic-content/components/SelectPostType.js
+++ b/src/extend/dynamic-content/components/SelectPostType.js
@@ -2,18 +2,24 @@ import { __ } from '@wordpress/i18n';
 import AdvancedSelect from '../../../components/advanced-select';
 import usePostTypes from '../hooks/usePostTypes';
 import { useMemo } from '@wordpress/element';
+import { applyFilters } from '@wordpress/hooks';
 
 export default ( { postType, onChange, value, help } ) => {
 	const postTypes = usePostTypes();
 
+	const enabledPostTypes = useMemo( () => {
+		const enabled = postTypes.filter( ( type ) => ( type.viewable && 'attachment' !== type.slug ) );
+
+		return applyFilters( 'generateBlocks.editor.query-loop.enabled-post-types', enabled, postTypes );
+	}, [ postTypes ] );
+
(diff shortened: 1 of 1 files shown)


src/extend/dynamic-content/components/SelectPostType.js
```diff
@@ -2,18 +2,24 @@ import { __ } from '@wordpress/i18n';
...
 import { useMemo } from '@wordpress/element';
+import { applyFilters } from '@wordpress/hooks';
 
 export default ( { postType, onChange, value, help } ) => {
 	const postTypes = usePostTypes();
 
+	const enabledPostTypes = useMemo( () => {
+		const enabled = postTypes.filter( ( type ) => ( type.viewable && 'attachment' !== type.slug ) );
+
+		return applyFilters( 'generateBlocks.editor.query-loop.enabled-post-types', enabled, postTypes );
```
tomusborne, line 13: I think our other filters use `generateblocks` instead of `generateBlocks`.

PR #669, opened 2022-08-23
Fix: Color picker behavior

### LINKED ISSUE #540: Jumping cursor in RGBA color pallete value editing

JeanPaiva:
closes #540


tomusborne (approved):
We need to make these changes to GB Pro as well (as well as GP).

PR #683, opened 2022-09-03
Fix: Missing CSS when do_blocks is ran early

tomusborne:
alternative to #682 

#682 works but has an issue where CSS (singular and block-specific) is duplicated when using block themes.

This method works by using `wp_enqueue_scripts` and `wp_add_inline_style` for our inline style. This works because `wp_enqueue_scripts` fires before `wp_head` which is where Slim SEO (in this case) is calling `do_blocks`.

We also have to check `$block_ids` within the `generateblocks_get_dynamic_css` function to prevent duplicate CSS in block themes.

Not sure if this is the best solution.

### DIFF
--- a/includes/class-enqueue-css.php
+++ b/includes/class-enqueue-css.php
@@ -65,7 +65,7 @@ class GenerateBlocks_Enqueue_CSS {
 		$dynamic_css_priority = apply_filters( 'generateblocks_dynamic_css_priority', 10 );
 
 		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_dynamic_css' ), $dynamic_css_priority );
-		add_action( 'wp_head', array( $this, 'print_inline_css' ), $dynamic_css_priority );
+		add_action( 'wp_enqueue_scripts', array( $this, 'print_inline_css' ), $dynamic_css_priority );
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


includes/functions.php
```diff
@@ -1056,7 +1056,7 @@ function generateblocks_get_dynamic_css( $content = '', $store_block_id_only = f
 				if ( is_callable( [ $blocks[ $name ], 'get_css_data' ] ) ) {
 					if ( $store_block_id_only ) {
 						$blocks[ $name ]::store_block_id( $atts['uniqueId'] );
-					} else {
+					} elseif ( ! in_array( $atts['uniqueId'], $blocks[ $name ]::$block_ids ) ) {
```
JeanPaiva, line 1059: Another option, without changing the attributes to public, is to create a method `$blocks[]::block_id_exists()`.

JeanPaiva (commented):
I could not find any CSS duplication with GP/2022 block theme, and widget areas and the original issue is still fixed.

includes/class-enqueue-css.php
```diff
@@ -161,8 +161,12 @@ public function print_inline_css() {
 				return;
 			}
 
-			printf(
-				'<style id="generateblocks-css">%s</style>',
+			// Add a "dummy" handle we can add inline styles to.
```
tomusborne, line 164: @JeanPaiva Not sure I like this dummy handle thing, but it's the only one I could find to add inline styles within the `wp_enqueue_scripts` hook: https://wordpress.stackexchange.com/a/282868

JeanPaiva (approved):

Issue #685, opened 2022-09-05
Button markup and accessibility


quantumleap33:
I think it would be a good idea to let the buttons have the `<button>` markup instead of `div` or `span`. It's better for accessibility and will come in handy in case the buttons are transformed into tab triggers for tab functionality.

Reference:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

PR #698, opened 2022-09-14
Feature: Allow Buttons outside of the Button Container block

tomusborne:
This allows the Button block to be added outside of the Button Container block.

Buttons within the Button Container will continue to use the `.gb-button-wrapper` selector to maintain backward-compatibility.

Buttons outside of the Button Container only use the `.gb-button` selector.

The only possible "breaking change" here is that the one-time CSS no longer has the wrapper selector.


quantumleap33:
What about adding the `<button>` markup to buttons as well?
https://github.com/tomusborne/generateblocks/issues/685

JeanPaiva (approved):

PR #701, opened 2022-09-23
Refactor inspector controls

JeanPaiva:
Refactoring inspector controls to be re-usable and customizable by all blocks.

- [x] #701
- [x] #704 
- [x] #707 
- [x] #710 
- [x] #711 
- [x] #714


src/blocks/headline/attributes.js
```diff
@@ -13,10 +13,6 @@ export default {
 		source: 'html',
 		selector: '.gb-headline-text',
 	},
-	element: {
```
tomusborne, line 16: This one is specific to the Headline block - we can likely keep it in here with the others.

JeanPaiva: Ideally, this control would be the same as the container "tagName" and they would share the same control component, changing only the options.

tomusborne (approved):

PR #707, opened 2022-09-29
Button block controls refactor

JeanPaiva:
Refactor Button block inspector controls


tomusborne:
This is working great. The only "issue" I'm seeing is the icon padding doesn't adjust when changing the icon location. It just removes the default padding instead of changing it over to the left/right.

JeanPaiva:
Fixed

tomusborne (approved):

PR #710, opened 2022-10-03
Refactor image block controls

JeanPaiva:
Image block controls refactor


src/block-context/image.js
```diff
@@ -0,0 +1,57 @@
...
+	id: 'image',
+	supports: {
+		responsiveTabs: true,
+		settingsPanel: {
+			enabled: true,
+			icon: 'backgrounds',
+		},
+		spacing: {
+			enabled: true,
+			inlineWidth: true,
```
tomusborne, line 15: This isn't an option in the Image block.

PR #715, opened 2022-10-10
Feature: New Layout controls

tomusborne:
- [x] #716 
- [x] #717 
- [x] #724


src/extend/inspector-control/controls/layout/components/LayoutControl.js
```diff
@@ -0,0 +1,58 @@
...
+import flexOptions from '../options';
+import './editor.scss';
+import classnames from 'classnames';
+
+export default ( props ) => {
+	const {
+		setAttributes,
+		label,
+		attributeName,
+		deviceType,
```
JeanPaiva, line 13: You could use directly the hook `useDeviceType` inside this component, instead of having to pass this down from every usage of this component.

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -0,0 +1,87 @@
...
+	return (
+		<PanelArea
+			title={ __( 'Layout', 'generateblocks' ) }
+			initialOpen={ false }
+			icon={ getIcon( 'layout' ) }
+			className="gblocks-panel-label"
+			id={ `${ id }Layout` }
+		>
+			{ layout.display &&
+				<SelectControl
```
JeanPaiva, line 31: You can create a new component for this one and pass only `value` and `onChange`, so we "hide" the static props.

src/extend/inspector-control/controls/layout/components/LayoutControl.js
```diff
@@ -0,0 +1,58 @@
...
+		attributes,
```
JeanPaiva, line 14: In order for this component to become stateless, you need to avoid passing all `attributes` list, any changes made to any of the attributes will trigger re-render here. You can pass the `isPrimary`, `onClick`, `direction`, and so on.

Maybe you can create one component for each "attributeName" and do all the logic there.

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -0,0 +1,87 @@
...
+							[ getAttribute( 'display', componentProps, true ) ]: value,
+						} );
+					} }
+				/>
+			}
+
+			{ isFlexItem( { device, attributes } ) &&
+				<>
+					{ layout.flexDirection &&
+						<LayoutControl
```
JeanPaiva, line 53: Maybe creating a component `FlexDirection` with the logic necessary for the LayoutControl to work is cleaner.

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -0,0 +1,87 @@
...
+							{ ...componentProps }
+							label={ __( 'Direction', 'generateblocks' ) }
+							attributeName="flexDirection"
+						/>
+					}
+
+					{ layout.alignItems &&
+						<LayoutControl
```
JeanPaiva, line 61: Same as above.

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -0,0 +1,87 @@
...
+							{ ...componentProps }
+							label={ __( 'Align Items', 'generateblocks' ) }
+							attributeName="alignItems"
+						/>
+					}
+
+					{ layout.justifyContent &&
+						<LayoutControl
```
JeanPaiva, line 69: Same as above.

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -0,0 +1,87 @@
...
+							{ ...componentProps }
+							label={ __( 'Justify Content', 'generateblocks' ) }
+							attributeName="justifyContent"
+						/>
+					}
+
+					{ layout.flexWrap &&
+						<LayoutControl
```
JeanPaiva, line 77: Same as above.

JeanPaiva (commented):
Made some suggestions, not necessarily required to do, but in order to keep consistency with other controls would be cool.

JeanPaiva (approved):

PR #720, opened 2022-10-11
Tweak: Move tagName option to Advanced

tomusborne:
This moves the Container tagName option to the Advanced panel.

It also removes the `containerAfterElementTag` filter as we're not using it.

### DIFF
--- a/src/blocks/container/components/TagName.js
+++ b/src/blocks/container/components/TagName.js
@@ -1,0 +1,38 @@
+import { SelectControl } from '@wordpress/components';
+import { applyFilters } from '@wordpress/hooks';
+import { __ } from '@wordpress/i18n';
+import { useDeviceType } from '../../../hooks';
+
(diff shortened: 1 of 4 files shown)


src/blocks/container/components/TagName.js
```diff
@@ -0,0 +1,38 @@
...
+export default function TagName( props ) {
+	const {
+		onChange,
+		tagName,
+	} = props;
+
+	const [ device ] = useDeviceType();
+
+	const tagNames = applyFilters(
+		'generateblocks.editor.containerTagNames',
```
tomusborne, line 15: Not sure what else we need in `props` here. Can't find us using this filter anywhere.

JeanPaiva: I think we just have this in case people wanted to have other tags in there. Don't think anybody is using it.

JeanPaiva (approved):

PR #727, opened 2022-10-17
Feature: Add flex child controls to Headline and Button blocks

tomusborne:
This adds flex child controls to the Button and Headline blocks.


JeanPaiva (approved):
This PR is adding many features for two blocks, still, the code is so simple. Loved it

PR #729, opened 2022-10-18
Feature: Add gap options to our blocks

tomusborne:
This adds `column-gap` and `row-gap` options to the Container, Button and Headline blocks.


JeanPaiva (approved):
Again, so easy to add new stuff.

PR #737, opened 2022-10-25
Feature: Add unit control presets

tomusborne:
This adds the ability to add presets to the `UnitControl` component when the field is focused.

The presets only display if the unit value matches the units in the preset values.

presets


src/extend/inspector-control/controls/sizing/components/MaxHeight.js
```diff
@@ -1,17 +1,14 @@
 import { __ } from '@wordpress/i18n';
 import UnitControl from '../../../../../components/unit-control';
 
-export default function MaxHeight( { value, desktopValue, tabletValue, onChange } ) {
+export default function MaxHeight( props ) {
```
JeanPaiva, line 4: Why switched to prop drilling here?

JeanPaiva (commented):
I believe this still does not solve the real issue, the user will not see that right away, he needs to click something to have some presets. I still think having a utility section with all "shortcuts" is a good thing. We can build that section based on the block context so it shows controls more tailored for the usage.

About showing the popover only for a specific set of containers (grid containers in this case) is something we must avoid. We should show that always or not show it at all.

src/extend/inspector-control/controls/sizing/components/MaxHeight.js
tomusborne: Seemed wasteful to add the props in both places consistently. Figured since they're getting a static set of props that it shouldn't matter. Probably just lazy lol

tomusborne:
I agree, I'm not sure this is the right way to handle this.

tomusborne:
@JeanPaiva should we keep the ability to do this but remove it from `Width` for now? We may want to tap into this kind of functionality in the future (font size etc..).

JeanPaiva (approved):

JeanPaiva:
@tomusborne yes, let's keep the functionality there for future use.

Issue #697, opened 2022-09-14
Dynamic data - missing editor buttons for bold / italic


onetrev:
### Description
If I enable Dynamic Data on a GenerateBlocks element, then suddenly the buttons for Bold and Italic disappear. A quick bit of CSS can fix overcome this, so not a major issue, but definitely not ideal.

### Steps to reproduce
1. Add a GB block to the WP editor.
2. Enable Dynamic Data

#### Notes: 
- This happens with latest GB Pro and with 1.4.0 Beta 1.

![Screenshot 2022-09-14 153212](https://user-images.githubusercontent.com/597601/190068483-9c009994-bb71-431f-abb6-20e5f6ee146d.jpg)

-----

tomusborne:
The reason for this is the bold/italic options add the necessary static HTML to the selected text as it's a Rich Text component.

Once you enable dynamic data, the block is no longer saved as static HTML. The content is generated dynamically on the frontend, so it's not possible to highlight specific words to bold/italicize them.

PR #739, opened 2022-10-25
Feature: Improve typography component

tomusborne:
This PR:

1. Adds the container text alignment options to the Typography panel.
2. Adds the headline text alignment options to the Typography panel
3. Fixes a bug where Google fonts were duplicated in the font family dropdown
4. Improves the UI of the typography panel.

typography2


JeanPaiva (commented):
Font family improvements are good.

Text alignment, why only for the containers? Even though we have it on inspectors' controls, I think having it in the toolbar is still necessary to keep consistency with other blocks.

Too much whitespace to the right, maybe having the label to the left could save some vertical space and remove that white box. Maybe we can add more controls there like bold, italic, and strikethrough?

Also, why don't we have "justify"?

tomusborne:
> Text alignment, why only for the containers? Even though we have it on inspectors' controls, I think having it in the toolbar is still necessary to keep consistency with other blocks.

We can/should do the same for Headline. Button doesn't have an alignment option, but it should. Moving from toolbar to inspector controls was discussed on Monday. David had mentioned it was confusing users as it wasn't obvious that the control was available for all devices.

Agree on whitespace, let me see if I can do something there.

Justify is a good one to add - any other alignments?

tomusborne:
@JeanPaiva 

1. Fixed the whitespace
2. Added justify
3. Kept the block toolbar controls

Will add alignment to the Button block in a separate PR as it involves some migration.

JeanPaiva (approved):

onetrev:
Hey @tomusborne ... I didn't want to create a new issue, but I noticed @JeanPaiva mentioned adding italic, strikethrough, etc. 

If this were added then you could have italics and strikethrough for example for **dynamic data blocks**.  You can't do with GenerateBlocks because there is no GUI option for it. But you can do that currently with WP Core Blocks, like the Post Date block, see screenshot below. 

Screenshot 2023-03-04 195910

I did a poor job of explaining [the issue](https://github.com/tomusborne/generateblocks/issues/697) previously. I understand you can't highlight specific words to be italic / strikethrough, but the whole block should be able to be italic or strikethrough right?

Happy to make a new issue for this if it's something you guys think you'd add one day. :)

PR #740, opened 2022-10-25
Tweak: Change panel icon color

tomusborne:
This makes the panel icons the same color as the text.

panel-icon


JeanPaiva (approved):
Personally, I didn't like it. But I don't have strong feelings about it

PR #743, opened 2022-10-26
Release: 1.7.0

tomusborne:
A game-changing release.

Still in active development.


eastriverstudio:
@tomusborne 

I hope you don't mind me jumping in here as I know active development is still taking place on the new 1.7.0 features which are looking great. A couple of early observations from my initial playing around with the new layout / sizing controls which are looking excellent by the way and exactly what we need to address some of the layout limitations we have within the block editor currently.

The main one is that with the new controls it still doesn't seem possible to easily push the last item (e.g a button) in a flex container to the bottom whilst everything else remains at the top. It may be possible be wrapping elements in extra cotnainers and setting the justification to space-between but in my mind that defeats the purpose of the simplified container. Therefore I'd like to see the option to add unitless values in the margin spacing controls. This would allow us to add the auto value which would easily allows us to push just a button block or another GB block element to the end in either flex direction.

The second one is I'd like to suggest a unitless option is added to the sizing panel. This would allow us to use the css calc() function to easily create a fixed sized element (e.g not percentage value) next to an element which fills the rest of space. I've found this highly useful in the existing grid using the flex-bias option with a unitless value using calc(). Whilst this may be possible with the new flex-child controls from my testing it would rely on extra wrapping containers where having this additional option in the sizing would again simplify I mean we could achieve that more simple without extra containers.

Finally still no rem option in the typography panel when it can be set in GP's typography panel globally. It would be nice to also be able to set rem in the block editor for consistency unless there's a reason this has not be added already?

I hope these suggestions make sense and if these elements I've mentioned are still WIP then do let me know.

Thanks

eastriverstudio:
Hi @tomusborne 

Ah ha, yes you can! To be honest I’d not even tried it on the sizing fields because I think I just saw the set unit options and assumed it wouldn’t work. As for the margin auto value, I did try that but it didn’t change anything on the editor side, but I can see it works on the frontend now. This is great news but agree it could be more obvious that these values are possible. 

Really excited by this release, it’s great to see these layout improvements coming to fruition. If you need an extra pair of hands to test anything specific on this release just let me know, happy to help where I can.

Thanks

eastriverstudio:
Also meant to add, have you guys considered adding options for positioning e.g absolute or fixed? Up to now I can see why positioning options haven't been available and I know it is a balancing act of giving enough control but not cluttering the UI with too many options to overwhelm. 

Now with the new container and sizing options being able to position certain elements outside of the usual page flow seems like it might be more realistic and useful propostion to incorporate, however I appreciate this might also throw up issues inside the block editor preview.

Thanks

src/hoc/withContainerLegacyMigration.js
```diff
@@ -16,6 +17,23 @@ export default ( WrappedComponent ) => {
 				setAttributes( { isDynamic: true } );
 			}
 
+			MigrateSizing( { attributes, setAttributes } );
```
tomusborne, line 20: @JeanPaiva Think we should move this into the `isBlockVersionLessThan` condition below?

includes/functions.php
```diff
@@ -1136,3 +1141,224 @@ function generateblocks_maybe_add_block_css( $content = '', $data = [] ) {
...
+	);
+}
+
+/**
+ * Returns the global $wp_filesystem with credentials set.
+ * Returns null in case of any errors.
+ *
+ * @return WP_Filesystem_Base|null
+ */
+function generateblocks_get_wp_filesystem() {
```
JeanPaiva, line 1329: We need an FTP server to test if connects and allows CSS file creation.

package.json
```diff
@@ -18,28 +18,28 @@
...
-		"@wordpress/data": "6.1.5",
-		"@wordpress/date": "4.2.3",
-		"@wordpress/element": "4.0.4",
-		"@wordpress/eslint-plugin": "10.0.1",
-		"@wordpress/hooks": "3.2.2",
-		"@wordpress/i18n": "4.2.4",
-		"@wordpress/icons": "6.1.1",
-		"@wordpress/rich-text": "5.0.7",
-		"@wordpress/scripts": "22.2.1",
+		"@wordpress/api-fetch": "6.13.0",
```
JeanPaiva, line 27: Maybe we can update these packages again?

readme.txt
```diff
@@ -85,6 +85,67 @@ GenerateBlocks was built to work hand-in-hand with [GeneratePress](https://gener
 
 == Changelog ==
 
+= 1.7.0 =
```
JeanPaiva, line 88: Epic changelog

src/hoc/withContainerLegacyMigration.js
JeanPaiva: https://github.com/tomusborne/generateblocks/pull/853

PR #775, opened 2022-11-18
Improve advanced select results

### LINKED ISSUE #732: Fix PostTypeRecordsSelect search results
### Description
The component `PostTypeRecordsSelect` is not working properly when the result set has more than 10 posts.

### Steps to reproduce
1. Have a site with 25+ posts (with similar words in the title)
2. Add a query loop
3. Add "Include posts" and search for a post that is not showing initially in the options.

#### Expected behavior
The search result brings all posts correctly

#### Actual behavior
The search result is only adding the first 10 posts from the result set.

JeanPaiva:
close #732

### DIFF
--- a/src/blocks/query-loop/components/inspector-controls/controls/TaxonomyParameterControl.js
+++ b/src/blocks/query-loop/components/inspector-controls/controls/TaxonomyParameterControl.js
@@ -80,7 +80,7 @@ export default function TaxonomyParameterControl( props ) {
 
 							setTerms( newTerms );
 						} }
-						help={ terms.length === 0 ? __( 'You must select at least one term. Use #123 with your ID to search by term ID.', 'generateblocks' ) : help }
+						help={ terms.length === 0 ? __( 'You must select at least one term. Search by name or ID.', 'generateblocks' ) : help }
 					/>
 
 					{ isHierarchical &&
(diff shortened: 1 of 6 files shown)


tomusborne:
@JeanPaiva Do we need to include this is the post select when querying/searching for specific posts in the Dynamic Content area?

JeanPaiva:
@tomusborne the posts select component is the same for both, the only thing missing was the label on the dynamic content.

tomusborne:
Playing with this. I think it may be better to:

1. Allow search by ID when using `#` or if the search is a number only.
2. Change the help text to: `Search by name or ID.`

What do you think?

JeanPaiva:
@tomusborne changed to accept only numbers as ID search and removed the # tag search.

src/blocks/query-loop/components/inspector-controls/controls/TaxonomyParameterControl.js
```diff
@@ -80,7 +80,7 @@ export default function TaxonomyParameterControl( props ) {
 
 							setTerms( newTerms );
 						} }
-						help={ terms.length === 0 ? __( 'You must select at least one term.', 'generateblocks' ) : '' }
+						help={ terms.length === 0 ? __( 'You must select at least one term. Use #123 with your ID to search by term ID.', 'generateblocks' ) : help }
```
tomusborne, line 83: @JeanPaiva Need to update help here.

PR #788, opened 2022-11-24
Fix: Container appender positioning

### LINKED ISSUE #776: Release 1.7 - block appender modal may fall outside viewport
### Description
New block appenders left hand alignment can lead to block list modal to fall outside of the viewport:

Screenshot 2022-11-21 at 15 53 06

tomusborne:
close #776 

This fixes #776 by adding our own version of the `ButtonBlockAppender` so we can tweak the `position` of the `Inserter` component.

Not ideal, but seems necessary for now until core fixes that issue.


JeanPaiva (approved):
Such a shame having all of that to change 1 prop.

PR #797, opened 2022-11-30
Feature: Add onboard system

JeanPaiva:
This adds an onboard popover for the Insert inner container toolbar button.

### DIFF
--- a/includes/class-rest.php
+++ b/includes/class-rest.php
@@ -36,6 +36,13 @@ class GenerateBlocks_Rest extends WP_REST_Controller {
 	protected $version = '1';
 
 	/**
+	 * Onboarding meta key.
+	 *
+	 * @var string
+	 */
+	const ONBOARDING_META_KEY = 'generateblocks_onboarding';
+
+	/**
 	 * Initiator.
 	 *
 	 * @return object initialized object of class.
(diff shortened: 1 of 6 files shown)


includes/class-rest.php
```diff
@@ -171,6 +181,38 @@ public function regenerate_css_files( WP_REST_Request $request ) {
...
+		$onboard = get_user_meta( $user_id, 'gb_onboard', true );
+		$key = $request->get_param( 'key' );
+
+		if ( ! $onboard ) {
+			$onboard = array();
+		}
+
+		$onboard[ $key ] = true;
+
+		update_user_meta( get_current_user_id(), 'gb_onboard', $onboard );
```
tomusborne, line 202: Let's name this `generateblocks_onboarding`

src/components/onboard-popover/index.js
```diff
@@ -0,0 +1,60 @@
...
+					focusOnMount={ true }
+					noArrow={ false }
+					variant="toolbar"
+					offset={ 10 }
+					flip={ true }
+				>
+					{ children }
+					<div className="gb-onboard-popover__button">
+						<Button
+							variant="tertiary"
```
tomusborne, line 37: Tertiary or Primary?

src/components/onboard-popover/onboards/insert-inner-container-onboard.js
```diff
@@ -0,0 +1,25 @@
...
+	return (
+		<OnboardPopover onboardKey="insert-inner-container">
+			<h3>
+				{ __( 'Insert inner container', 'generateblocks' ) }
+			</h3>
+			<p>
+				{ __( 'Use this button to add a centered inner Container block for your content.', 'generateblocks' ) }
+			</p>
+			<p>
+				{ __( 'We generally recommend this for top-level Containers so your content doesn\'t span the entire width of your screen. ', 'generateblocks' ) }
```
tomusborne, line 14: Let's use `does not` instead of `doesn't` so we don't need to escape it.

tomusborne (approved):

PR #868, opened 2023-01-16
Fix: Grid columns dont have width set correctly

### LINKED ISSUE #863: 1.7 Grid Block, the initial grid columns have no widths
### Description
When adding a new Grid and selecting an option from the Select Layout panel, the grid items get loaded without any widths:

https://user-images.githubusercontent.com/30641942/212466869-80a25c23-8a94-4662-bddb-b1fdcf8de959.mov

JeanPaiva:
Closes #863

### DIFF
--- a/src/blocks/grid/components/LayoutSelector.js
+++ b/src/blocks/grid/components/LayoutSelector.js
@@ -28,6 +28,9 @@ export const getColumnsFromLayout = ( layout, uniqueId ) => {
 		paddingLeft: gridItemPaddingLeft || '',
 		widthMobile: 100,
 		width: Number( col ),
+		sizing: {
+			width: `${ col }%`,
+		},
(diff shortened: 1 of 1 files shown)


src/blocks/grid/components/LayoutSelector.js
```diff
@@ -28,6 +28,9 @@ export const getColumnsFromLayout = ( layout, uniqueId ) => {
 		paddingLeft: gridItemPaddingLeft || '',
 		widthMobile: 100,
 		width: Number( col ),
```
tomusborne, line 30: I think we can remove this.

JeanPaiva: @tomusborne done, removed the width.

tomusborne (approved):

PR #869, opened 2023-01-16
Feat: Add filter for the image block link attributes

JeanPaiva:
Add filter to allow customization for the image block dynamic link anchor tag attributes.


tomusborne:
@JeanPaiva I tweaked this so it uses `generateblocks_attr()` as that's the standard with our other HTML elements.

PR #880, opened 2023-01-25
Fix: Static image caption showing multiple times

JeanPaiva:
References:
https://secure.helpscout.net/conversation/2134167229/373534
https://generatepress.com/forums/topic/the-caption-in-the-image-block-appears-3-times/#post-2506778


tomusborne:
@JeanPaiva This breaks our blocks that use static content with a dynamic link.

PR #887, opened 2023-01-26
Fix: Missing responsive controls when multi-selecting blocks

### LINKED ISSUE #885: Ability to select responsive icons when multiple blocks are selected
Is it possible to add the ability to select responsive icons when multiple blocks are selected from the list view? It would help with multiple block  setting adjustments in the various responsive views.
https://www.loom.com/share/73381a5bbb5a4df5a2d8f12eb3c75172

tomusborne:
close #885


tomusborne:
@JeanPaiva Don't think updating dist files should have happened here.

PR #915, opened 2023-02-20
Fix: Display option changing to flex when changing icon

tomusborne:
Not sure if this is crucial enough to merge into 1.7 or if it should wait.


JeanPaiva:
It works well, but if for some reason the user wants the display to be something else the issue is still there. Maybe we check if the icon was already set and don't apply any other rule. But I think this can wait for the next version.

tomusborne:
This is being handled in #983

PR #954, opened 2023-03-27
Chore: Add PHP tests

JeanPaiva:
Add unit tests to the plugin.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-versions: ['5.6', '7.4', '8.1']
+        php-versions: ['7.2', '7.4', '8.1']
     steps:
       - name: Checkout repo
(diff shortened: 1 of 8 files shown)


.env.testing-sample
tomusborne, line 1: Does this mean anyone testing has to set up this same environment?

.github/workflows/ci.yml
```diff
@@ -139,7 +139,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-versions: ['5.6', '7.4', '8.1']
```
tomusborne, line 142: No more 5.6 testing?

.gitignore
tomusborne, line 1: We'll need to update `.distignore` with all newly-added files/folders as well.

.github/workflows/ci.yml
JeanPaiva: Yes, the tests require at least PHP 7.2

tomusborne (approved):

PR #957, opened 2023-03-27
Feat: Display text field to allow user to replace image URL

### LINKED ISSUE #952: QoL - Image Block improvements
### Description
1. Improve Replace Button so user can see a URL if one was added.

JeanPaiva:
Closes #952


JeanPaiva:
@tomusborne updated to not show if mediaId is set.

tomusborne (approved):

Issue #938, opened 2023-03-05
Duplication in widgets does not generate new ID for GB blocks


leohsiang:
Normal content (correct) behavior when the 2nd and 3rd items are duplicated:
https://www.screencast.com/t/zlE3j0Vy
https://www.screencast.com/t/YOMx2v3vlG

Widget area (incorrect) behavior when the 2nd and 3rd items are duplicated:
https://www.screencast.com/t/YJAeKuiEtE
https://www.screencast.com/t/5meIfrYlfx5c

JeanPaiva:
Related to https://github.com/tomusborne/generateblocks/issues/129

PR #958, opened 2023-03-27
Fix: unique id not regenareted correctly on widgets editor

### LINKED ISSUE #129: Add compatibility with the widget editor
### Description
`getBlocks()` doesn't work in the widget editor. We should look at adding compatibility with the widget editor (or figuring out a different way to know all the unique IDs being used).

### Steps to reproduce
1. 

#### Expected behavior
The logic to check if we have uniqueIds should match `innerBlocks`.

#### Actual behavior
uniqueId from `innerBlocks` is not returned from the function.

#### References: 
WordPress/gutenberg#32607

JeanPaiva:
Closes #129 #938

### DIFF
--- a/src/hoc/test/withUniqueId.test.js
+++ b/src/hoc/test/withUniqueId.test.js
@@ -1,0 +1,125 @@
+import getEditorBlocks from '../../utils/get-editor-blocks';
+import { getUniqueIdFromBlocks, hasDuplicates } from '../withUniqueId';
+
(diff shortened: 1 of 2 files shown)


tomusborne (approved):
Nice one. Feels like we could use a unit test or two here.

JeanPaiva:
@tomusborne added tests for the uniqueID internals.

PR #962, opened 2023-03-29
Feature: Add block appender toolbar

tomusborne:
close #944 

This feature adds a handful of new appenders to our block toolbar.

You can now:

* "Unwrap" a Container block, removing it and "un-containing" its inner blocks.
* Use the "Add to Container" block to wrap any block in the editor in a Container block.
* Use the "Add to Row" or "Add to Stack" buttons to add any multi-selection of blocks to a flex row or column.

We also moved some existing toolbar buttons into this new toolbar section.

* The "Add Inner Container" button appears in root-level Container blocks.
* The "Add Button" buttons appear in this section for the Button block and Button Container block variation.


JeanPaiva:
@tomusborne Accordions/Tags are breaking, just add them to the page and select the block.

Screenshot 2023-04-06 at 15 29 22

JeanPaiva:
When selecting multiple blocks, we can't add our own "grouping" function, but we need to allow users to transform from a Group block to a Container block. I think it's a good way of handling that limitation.

The user selects all blocks, groups them, and transforms them to Container.
Screenshot 2023-04-06 at 15 33 31

tomusborne:
Accordions/tabs should be fixed.

Will look at a Group -> Container transform, good idea.

JeanPaiva (approved):

PR #986, opened 2023-04-28
Tweak: Align buttons in a row using alignment matrix

tomusborne:
This sets the `flexDirection` to `row` when using the Alignment Matrix with nothing but the Button block.

This is only necessary if we decide to ditch the Button Container variant.


JeanPaiva:
Why is this needed? The "Add to Row" toolbar button already sets the container to flex row direction.

PR #987, opened 2023-04-28
Tweak: Remove the Button Container variation

tomusborne:
blocked #986 

There's an argument to be made that the "Button Container" variation is no longer necessary with the new Container block and the ability to easily wrap blocks inside of it. This PR explores that.


JeanPaiva:
Why not keep the Add to row/column buttons for the toolbar?

PR #1011, opened 2023-05-19
Fix: Migrate global style attributes in the editor

tomusborne:
This runs our Global Style attributes through our migration functions to ensure blocks using old Global Styles appear properly in the editor.


JeanPaiva (commented):
Found a bug, when borders are linked, the border hover color sets the primary color of the other sides if you unlink.

Screenshot 2023-05-23 at 12 50 50

JeanPaiva (approved):

PR #1025, opened 2023-05-31
Feature: Only show one dimension field if synced

tomusborne:
This makes it so only one dimension field displays if all 4 sides are synced.

### DIFF
--- a/src/components/dimensions/index.js
+++ b/src/components/dimensions/index.js
@@ -32,7 +32,9 @@ export default function Dimensions( props ) {
 		return o;
 	}, {} );
 
-	const [ sync, setSync ] = useState( false );
+	const uniqueValues = Object.values( attributes ).filter( ( n ) => n );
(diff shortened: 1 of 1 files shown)


JeanPaiva (approved):
One thing that I noticed is that when you are on the tablet device and change to mobile, the sync persists. Not sure if it was intentional or not. Apart from that all good.

tomusborne:
Hmm, I would say not intentional. Tried to make it so it only auto-syncs if the values all exist and are the same. Seems the component isn't re-rendering from tablet -> mobile.

tomusborne:
@JeanPaiva what about the latest commit?

JeanPaiva:
Another one:

When syncing on the desktop and going to the tablet, the placeholder is only showing for the top side.

Screenshot 2023-06-02 at 12 15 01

tomusborne:
@JeanPaiva this is using the latest pull?

JeanPaiva:
@tomusborne all good now.

PR #1026, opened 2023-05-31
Feature: Sync all border values on sync click

tomusborne:


JeanPaiva (changes requested):
Found two bugs.

Steps to reproduce the first:

1. Add width values to all 4 borders and remove them.
2. Save the page, and refresh, the sync is now active.

Steps to reproduce the second:

1. Add border width, style, and color, to any side except the TOP side.
2. Enable sync, click on the color field, no color is set.
3. Un-sync, the same occurs only for the top side, all other sides work fine.

tomusborne:
@JeanPaiva first bug should be fixed. Any ideas on what's going on with the second one? Value looks like it exists, maybe the component just isn't re-rendering?

JeanPaiva:
@tomusborne I fixed the second issue.

JeanPaiva (approved):

PR #1098, opened 2023-08-09
`wp_img_tag_add_loading_attr` deprecated in 6.3

danmaby:
`wp_img_tag_add_loading_attr` has been deprecated in WordPress 6.3. updated with `wp_img_tag_add_loading_optimization_attrs` https://developer.wordpress.org/reference/functions/wp_img_tag_add_loading_optimization_attrs/


JeanPaiva:
Your solution is good, but you need to add the `function_exists` check in order to not break websites using older versions of WP.

I'm closing this PR in favor of #1099 as it follows our internal workflow.

Thank you for your contribution.

PR #1130, opened 2023-10-18
Feature/dist files

dgwyer:
(This is mainly a test PR to make sure I'm doing it correctly.)

Removing the dist folder from the repo, as this should only be dynamically built during a release? Is that right? Chose to merge into the `feature/pattern-library` branch as that is the most active right now. Hope that's ok.


dgwyer:
Not needed. dist folder should always be present.

PR #1132, opened 2023-10-18
Release: 1.9.0

tomusborne:


tdrayson:
Is there any fuzzy search capabilities for searching class names?

For example if I have a class called `.container-width-l` searching for `width` in the box doesn't bring up any results.

tomusborne:
> Is there any fuzzy search capabilities for searching class names?
> 
> For example if I have a class called `.container-width-l` searching for `width` in the box doesn't bring up any results.

This would be a cool feature to add. Thanks!

PR #1142, opened 2023-11-03
Tweak: Change migration deps

tomusborne:
Related: https://github.com/tomusborne/generateblocks-pro/issues/350

We need to be able to re-fire these migrations if necessary.


tomusborne:
No longer needed thanks to classes re-write.

PR #1155, opened 2023-11-29
Feature/pattern library requests

### LINKED ISSUE #1153: Optimize Pattern Library Searching
### Description
The pattern library search currently re-fetches requests on every search and changing categories feels slow. My thinking is that we can just load all of the patterns initially and then filter them client side to speed up the interaction.

We may need to test this out to see how it works with a large number of patterns. I can think of a few potential footguns here, namely the size of the request. Currently the request payload is beefy, coming in at around _427kb_ and 9 total requests just to open up the modal. 

We need to streamline this to reduce the number of network requests and see what we can reasonably do to offload filtering and searching to the client, which should become progressively simpler as the filtered items are already lazy loaded themselves. 

## Technical Goals
- Reduce unnecessary re-renders and network requests
- Cache searches client side if possible (ex: using object key/value model)
- Eliminate multiple network requests when filtering categories or performing searches
- Consider implementing the new autocomplete component for searching
- Look at ways we can avoid a massive network request when there's a large number of patterns

## UX Goals
- Ensure users can quickly search and filter patterns
- Ensure the search is as meaningful as it can be with our available data

iansvo:
Fixes #1153

**Changes**
- Updated iframe attributes to help ensure smoother loading
- Eliminated extra fetch when setting library categories. All patterns for the library are loaded up front and then they are filtered client side. 
- Updated items_per_page to 15 to match new max row count
- Memoized category list to prevent extra re-rendering
- Added client side filtering logic and a local search cache. This will help the search feel snappier as it gets used (while the modal is open). 
- Various tweaks and minor style bugfixes

**Notes**
- I think that we should look towards a list virtualization solution. This would be a larger project of its own but it would allow the pattern library to handle a larger amount of patterns even better than it does now. 
- At 187 patterns, the performance is definitely not poor, but I think it could be better with optimizations such as list virtualization (as noted above). 
- Many pro patterns have large images (> 100kb) which should be optimized (noted in discord)
- I think we should consider moving toward using React query to clean up the useEffect soup in some of this (and other) code. I didn't think that made sense to try here as the life would've exceeded the value with our team's priorities.

### DIFF
--- a/src/blocks/pattern-library/components/library-layout.js
+++ b/src/blocks/pattern-library/components/library-layout.js
@@ -3,7 +3,7 @@ import { close, arrowLeft } from '@wordpress/icons';
 import { useDispatch } from '@wordpress/data';
 import { store as blockEditorStore } from '@wordpress/block-editor';
 import { __ } from '@wordpress/i18n';
-import { useState } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import CategoryList from './category-list';
 import LibrarySelector from './library-selector';
 import { useLibrary } from './library-provider';
@@ -14,6 +14,8 @@ import { PatternDetailsHeader } from './pattern-details-header';
 import RequiredComponents from './required-components';
 import LibraryCache from './library-cache';
 import ManageLibraries from './manage-libraries';
+
+const searchCache = {};
 
 export default function LibraryLayout() {
 	const {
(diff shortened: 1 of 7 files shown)


src/blocks/pattern-library/components/library-layout.js
```diff
@@ -94,19 +140,37 @@ export default function LibraryLayout() {
 				<div className="gb-pattern-library__sidebar">
 					{ ! activePatternId &&
 					<>
-						<PatternSearch />
-						<CategoryList />
+						<PatternSearch onChange={ ( value ) => {
```
iansvo, line 143: @tomusborne This is where I set up this simple AF cache logic for search I've mentioned before.

src/blocks/pattern-library/components/library-provider.js
```diff
@@ -180,8 +179,8 @@ export function LibraryProvider( { clientId, children } ) {
 			setRequiredClasses( [] );
 		}
 
-		// Fetch patterns for the active library.
-		const { data: fetchedPatterns } = await fetchLibraryPatterns( activeLibrary?.id, activeCategory, search, isLocal, publicKey );
+		// Fetch all patterns for the active library.
+		const { data: fetchedPatterns } = await fetchLibraryPatterns( activeLibrary?.id, '', '', isLocal, publicKey );
```
iansvo, line 183: As noted previously this is being left in place so we can use that REST endpoint to fetch a subset of data in the future.

tomusborne (approved):
Works great, looks good to merge!

PR #1156, opened 2023-12-07
Feature/styles indicator

iansvo:
Changes required for the styles indicator to work. 

### Relates to 

Issue: https://github.com/tomusborne/generateblocks-pro/issues/377
PR:


src/components/panel-area/index.js
```diff
@@ -33,42 +36,61 @@ export default function PanelArea( props ) {
...
-						const isOpen = panels[ id ] ||
+		<ApplyFilters
+			name="generateblocks.editor.panel"
+			blockName={ blockName }
+			state={ state }
+			panelRef={ ref }
+			{ ...props }
+		>
+			<ApplyFilters
+				name={ `generateblocks.editor.panel.${ blockName }${ capitalizeFirstLetter( id ) }` }
```
tomusborne, line 52: Are we sure this is the same as the previous `'generateblocks.panel.' + id` value?

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -41,227 +66,318 @@ export default function Layout( { attributes, setAttributes } ) {
...
+			hasGlobalStyle={ hasGlobalStyle }
 		>
 			{ !! useInnerContainer &&
-				<LegacyLayoutControls
-					attributes={ attributes }
-					setAttributes={ setAttributes }
-					deviceType={ device }
-					blockDefaults={ generateBlocksDefaults.container }
-				/>
+			<LegacyLayoutControls
```
tomusborne, line 160: Is there a reason for the no tabs here? Maybe I've been doing it wrong lol

src/components/panel-area/index.js
tomusborne: Related: #1161

src/extend/inspector-control/controls/spacing/index.js
```diff
@@ -48,14 +139,18 @@ export default function Spacing( { attributes, setAttributes, computedStyles } )
...
 					label={ __( 'Margin', 'generateblocks' ) }
 					attributeNames={ marginAttributes }
 					values={ deviceAttributes.spacing }
 					placeholders={ marginAttributes.reduce( ( o, key ) => (
-						{ ...o, [ key ]: getResponsivePlaceholder( key, attributes.spacing, device, computedStyles[ key ] ) }
+						{ ...o, [ key ]: getResponsivePlaceholder(
+							key,
+							attributes.spacing,
+							device,
+							parseInt( computedStyles[ key ] ) || ''
```
tomusborne, line 152: Not sure what's up here, but I'm seeing some weirdness with the placeholder.


https://github.com/tomusborne/generateblocks/assets/20714883/44f1c99b-1a71-4bdb-be7c-4042d6266b6d

tomusborne: Same thing happens with the typography font size field.

src/components/panel-area/index.js
iansvo: Hmmm yeah I think if you've changed those IDs back there's no reason to do them like this anymore.

src/extend/inspector-control/controls/layout/index.js
iansvo: Hmmm sorry about that, that must've been some kind of automatic lint fix, I didn't do that on purpose (I don't like that kind of formatting anyway) so I just changed it back to have another tab in there.

src/extend/inspector-control/controls/spacing/index.js
iansvo: This appears to be a sort of race condition where the final value before someone is done typing is the last thing rendered there. It's probably worth having some kind of onBlur handler to reset that when someone tabs away or something along those lines, which wouldn't be super hard to rig up. I'll see about that today.

iansvo: Resolved by including the placeholder prop as an effect prop for the UnitControl which seems to resolve this issue. 

I believe the reason this was happening previously is because the computedStyle object we're passing is an actual computed styles object, vs just an object that holds the value. So it updates faster/differently than the previous state value we used instead, etc.

PR #1158, opened 2023-12-12
Tweak: Add `useStyleIndicator` hook

tomusborne:
@iansvo Was looking at what it would take to extend this to the other panels and saw an opportunity to reduce duplication.

Should allow us to keep this part of the code in one place in case we need to bug fix/improve.


iansvo (approved):
This looks great, I've been so focussed on just getting it to work I didn't take the time to look at stuff like this. I dig it!

PR #1167, opened 2024-01-04
Fix: Pattern previews not loading

tomusborne:
The conditional `src` loading was causing previews not to load once we switched to loading core patterns.


iansvo (approved):
This is a big change I hope you're sure!

PR #1168, opened 2024-01-04
Update SortableList

iansvo:
Updates to match the changes in pro


iansvo:
Went ahead and merged this after some simple tests, code has all been tested pretty thoroughly and everything is supposed to be the same in both anyway. 

Eventually this can be replaced with our @edge22/components package!

PR #1171, opened 2024-01-17
Feature: De-blockify pattern library

tomusborne:
This removes the new Pattern Library as a block, and adds it as a modal you can toggle in the block editor toolbar.

https://github.com/tomusborne/generateblocks/assets/20714883/ddfb66eb-a938-4014-aa0e-81e900fd5a65


dgwyer:
Just tested this and had a weird issue where I couldn't get the pattern library button to display. Turns out that the latest update to the Gutenberg plugin changes the toolbar markup!

I added a fix for this to display the pattern library button using the new selector but there's a weird double render of the button. On page load it appears briefly, disappears, and then reappears. I assume that this is a side-effect of having the Gutenberg plugin active but there's no way to know until the new toolbar markup is part of core.

To make matters worse when the Gutenberg plugin is _not_ active and the collapse toolbar UI is displayed it messes with the position of the button. The collapsible button has a fixed location which is really annoying if you want to append UI to the toolbar.

### Toolbar expanded

image

### Toolbar collapsed

image

I added a fix for this too.

When Gutenber is active this isn't an issue. Looks like they've remove the expand/collapse toolbar UI with the latest changes. However, we need to keep in this fix for a while for existing sites. It's a real pain when the UI is changed like this, we just need to try and keep an eye on future changes to make sure nothing breaks.

I also played around with making the button more prominent as some competitors do. Also, I think we probably need to add some branding to the pattern library button (logo?), or inside the modal, or both. Users may not know otherwise that this is the GenerateBlocks pattern library. Especially if they install a page builder plugin which also appends a pattern library button to the toolbar.

tomusborne:
It's a shame we have to rely on existing markup to add this button. I wonder if there's a better place with proper support that would be obvious for the user and convenient to click to get to the patterns quickly.

SchneiderSam:
As a GB user i prefer something Not To prominent.

tomusborne:
I think this is more elegant (and not in the way). It also doesn't care about the full screen/toolbar configuration.


https://github.com/tomusborne/generateblocks/assets/20714883/69484058-41f8-4dec-9318-bc3c4e5cd8cc

Stormrunn3r:
Maybe a unconventional idea. But couldn't the GB Patterns be added to the WordPress pattern library by use of `register_block_pattern`? WordPress already does this for their own core patterns in `_load_remote_block_patterns`

I would have it work something like the following:

- On install request the GB Patterns from the generateblocks.com endpoint.
- Save the response as JSON inside a protected directory (not publicly accessible).
- Use `register_block_pattern` to register the patterns in the JSON file.

This would probably also be more performant as the function for WordPress does a GET request on every load of Gutenberg see [`_load_remote_block_patterns`:218](https://github.com/WordPress/wordpress-develop/blob/6.4/src/wp-includes/block-patterns.php#L218)

nickfmc:
I like @Stormrunn3r idea, the core WP patterns work great now, and I've found that explaining to an end user not to use the button that is right there on your block that says "create pattern" and use the Generate blocks one instead is super confusing for them. It sounds like https://github.com/tomusborne/generateblocks/pull/1143 will at least allow us to use GB containers inside core patterns now and have 2 on the same page not conflict, so that's Awesome if I read that one correctly.

If GB's blocks fully work inside of WP core Pattern library with fix 1143 then I'd at least like an option in the GB settings to hide the GB pattern library button from the editor, so end users don't have access to it directly.

personally I would love to see the GB local patterns go away completely and then the name of the Pattern Library changing to get rid of the duplicate term of pattern and replace it with something like "GB layouts" "GB pre-made Sections" something to eliminate user confusion. this is one of my biggest gripes with client training right now is how to use/save re-usable sections. to the point where I just tell them "there are bugs, just copy and paste the section you want each time"

tomusborne:
Thanks for all the feedback!

This is very similar to the direction we're going in.

The next versions of GB + GB Pro deprecates our current Local Patterns feature/post type. Instead, we've built a library based on the core Pattern post type. This means you can use core features to create a pattern, or sync (and partially sync in the future).

Our library will:

1. Connect you to our free and pro remote libraries
2. Show your local patterns that can be put into collections and categories
3. Share these collections across sites using our Cloud feature (in the near future)

PR #1175, opened 2024-01-30
Moved GB patterns button to the left

dgwyer:
Fixes [#447](https://github.com/tomusborne/generateblocks-pro/issues/447)

image

### DIFF
--- a/src/pattern-library/index.js
+++ b/src/pattern-library/index.js
@@ -7,14 +7,25 @@ import { __ } from '@wordpress/i18n';
 import './editor.scss';
 import getIcon from '../utils/get-icon';
 
+// Create a new element for the portal content here so it's only created once.
+const portalElement = document.createElement( 'div' );
+
 function PatternLibrary() {
 	const [ isOpen, setIsOpen ] = useState( false );
-	const toolbar = document.querySelector( '.interface-pinned-items' );
+	const container = document.querySelector( '.edit-post-header__settings' );
+
+	if ( ! container ) {
+		return null;
+	}
+
+	// Insert the new element at the start of the container
+	container.insertBefore( portalElement, container.firstChild );
 
 	return (
 		<>
-			{ !! toolbar && createPortal(
+			{ !! container && createPortal(
 				<Button
+					style={ { paddingRight: 'start' } }
 					className="gblocks-pattern-library-button"
 					onClick={ () => setIsOpen( true ) }
 					icon={ getIcon( 'generateblocks' ) }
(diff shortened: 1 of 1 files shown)


tomusborne:
Definitely better/not as jumpy vs being on the right.

Did up a quick alternative #1176. I like your PR as it's a little more obvious, but I like that mine doesn't rely on existing markup that could change at any time.

I'm 50/50 split, going to leave it up to everyone else lol

dgwyer:
> Definitely better/not as jumpy vs being on the right.
> 
> Did up a quick alternative #1176. I like your PR as it's a little more obvious, but I like that mine doesn't rely on existing markup that could change at any time.
> 
> I'm 50/50 split, going to leave it up to everyone else lol

I know what you mean but I don't think it's a big issue (IMO) if the markup changes. We can easily test with the Gutenberg plugin active to get an advanced warning of any changes so we can update it long before it would be an issue for any of our users.

Re your PR. I had this thought too, about making it just an icon. I considered that this could be a good approach if we decide to make a GB 'panel' (perhaps just use the GB logo for the icon), but for direct access to the patterns modal I think moving the current button to the left works best for me. 🙂

@diggeddy @iansvo Thoughts?

diggeddy:
personally I prefer the icon button to the right,
placing it to the left eats into (top bar) block toolbar space and users have reported bizarre z-index issues with that area; that concerns me.

I think we should leave the icon to the right for now. And look at how that can become a GB panel menu. Today it could house Patterns and Styles.

tomusborne:
I started on a PR that uses a general GB sidebar, but it seemed weird having an entire section with a single button labeled "Launch Pattern Library".

That being said, this could make it way more clear that this is indeed the GB pattern library vs something else.

dgwyer:
Some more ideas...

Having a button that opens a sidebar would be the most stable solution. I'd favor this just being the current button. i.e. the GB logo but without the 'Patterns' text.

There is a slight downside to this as it means users will have to click _twice_ to open the patterns modal which isn't ideal. However, we could leverage this by adding more items to the sidebar.

Maybe include quick start items, such as showing the last several patterns chosen; and perhaps other items (needs further thought - be able to select the default pattern library or something). Think of the GB Patterns sidebar as a _mini-patterns library_ so that users can quickly navigate to it and perform common operations and then open the full modal if necessary?

I don't think anyone else is doing anything like that right now so could be unique. Of course, this could be worked on in the next release and for now just have a mostly blank sidebar to act as launcher. We could just add links to docs, and some text info etc. so it doesn't look too bare for now.

PR #1176, opened 2024-01-30
Feature: Add pattern library to plugin sidebar

tomusborne:
Alternative to #1175 

https://github.com/tomusborne/generateblocks/assets/20714883/d37d2a4f-3f03-471e-8cc5-c64a875c25be

Did this as a quick alternative to using a portal to insert the Pattern Library block.

This has the advantage of using core features to add the toggle (no portal needed).

However, it has the disadvantage of not being as obvious/visible.

Nice thing here is that if markup changes, the button will remain there and it will open a regular plugin sidebar with a button to open the library.

### DIFF
--- a/src/pattern-library/components/library-layout.js
+++ b/src/pattern-library/components/library-layout.js
@@ -16,7 +16,7 @@ import { doAction } from '@wordpress/hooks';
 
 const searchCache = {};
 
-export default function LibraryLayout( { setIsOpen } ) {
+export default function LibraryLayout( { closeModal } ) {
 	const {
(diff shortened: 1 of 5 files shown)


dgwyer:
Looks like folks (from GB Discord channel) are favoring this approach over #1175 so we should probably go with this version. Only other minor point is if we stick with the current icon or swap it out for the GB icon?

tomusborne:
Yea, the icon is something we need to figure out. Will get a few ideas on it. Let's merge this for now anyway.

PR #1178, opened 2024-02-01
Fix: Wrong const

### LINKED ISSUE #1177: WSOD when GB Pro not active
Just came across this during testing. Anyone else seeing this when GB is active but GB Pro isn't?

image

@tomusborne Looks like this was introduced in this [commit](https://github.com/tomusborne/generateblocks/commit/d4d8f3e3c8d3bc0c5a736546e91aa79be2195520) in `/incluses/functions/php`.

Is this a typo? i.e. Should it be `GENERATEBLOCKS_DIR`?

tomusborne:
closes #1177

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1798,7 +1798,7 @@ function generateblocks_get_enqueue_assets(
 		return $fallback_assets;
 	}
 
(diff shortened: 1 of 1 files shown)


dgwyer:
Ok to merge this straight in and I'll retest to confirm?

PR #1181, opened 2024-02-08
Add support for generative UI to be added in the pro version

dgwyer:
Adds a slot for the generative AI UI to be added in GB Pro.


dgwyer:
Closing as the generative UI is now located in the GB panel.

PR #1190, opened 2024-02-14
Security: Add user capability check to Query Loop post status

tomusborne:
closes https://github.com/tomusborne/generateblocks-pro/issues/485

This prevents users without the proper capability from seeing posts with any status other than `publish` in the Query Loop in the editor and frontend.

### DIFF
--- a/includes/class-query-loop.php
+++ b/includes/class-query-loop.php
@@ -129,6 +129,15 @@ class GenerateBlocks_Query_Loop {
 			$query_args['posts_per_page'] = $per_page;
 		}
 
+		if (
+			isset( $query_args['post_status'] ) &&
+			'publish' !== $query_args['post_status'] &&
+			! current_user_can( 'read_private_posts' )
+		) {
+			// If the user can't read private posts, we'll force the post status to be public.
+			$query_args['post_status'] = 'publish';
+		}
+
 		return $query_args;
 	}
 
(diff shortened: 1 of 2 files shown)


includes/class-query-loop.php
```diff
@@ -129,6 +129,15 @@ public static function get_query_args( $block, $page ) {
 			$query_args['posts_per_page'] = $per_page;
 		}
 
+		if (
+			isset( $query_args['post_status'] ) &&
+			'publish' !== $query_args['post_status'] &&
+			! current_user_can( 'read_private_posts' )
```
iansvo, line 135: Is there any permission check needed for drafts? Seems like this is only really stopping someone from seeing private posts.

tomusborne:
> Should the UI even allow someone to set a post_status they aren't allowed to query?

Thought this as well, but someone could just add the status to the block attributes via the code editor.

includes/class-query-loop.php
tomusborne: Hmm, I'll see if there's a better capability we can use here.

iansvo:
I suppose both things seem relevant, right? You'd want to stop the manual edit as well as just holster the footgun of letting someone pick a status they can't query (arguably more UX than security).

tomusborne:
It's super difficult (not possible?) to prevent people from adding their stuff into the code editor.

This solution should prevent anything bad, and I believe `read_private_posts` applies to drafts as well as far as I'm reading, although it's not super clear.

dgwyer:
Tested. Fix works locally for me!

PR #1226, opened 2024-04-15
Feature: CSS Grid

tomusborne:
This PR introduces a CSS Grid feature to our local block styles.

It also adds a new "Grid" variation of the Container block that replaces our current "Grid" block.


src/blocks/container/components/GridBlockSettings.js
```diff
@@ -0,0 +1,83 @@
...
+					<TextControl
+						id="grid-template-columns"
+						value={ deviceAttributes.gridTemplateColumns }
+						onChange={ ( value ) => setDeviceAttributes( { gridTemplateColumns: value } ) }
+					/>
+					<Button
+						size="small"
+						onClick={ () => setShowGridTemplateColumns( ! showGridTemplateColumns ) }
+						icon={ () => <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 256"><rect width="256" height="256" fill="none" /><rect x="32" y="56" width="192" height="144" rx="8" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="96" y1="56" x2="96" y2="200" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="160" y1="56" x2="160" y2="200" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="32" y1="104" x2="224" y2="104" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="32" y1="152" x2="224" y2="152" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /></svg> }
```
iansvo, line 39: Should we perhaps use getIcon here?

src/blocks/container/templates.js
```diff
@@ -0,0 +1,198 @@
+import { addFilter } from '@wordpress/hooks';
+import { __ } from '@wordpress/i18n';
+import classnames from 'classnames';
+
+function buildLayout( layouts, layout ) {
+	return (
+		<div
+			label={ layout }
```
iansvo, line 8: Is that supposed to be `aria-label`? That doesn't appear to be a valid attribute.

src/blocks/container/templates.js
```diff
@@ -0,0 +1,198 @@
...
+	for ( let i = 0; i < repeatCount; i++ ) {
+		innerBlocks.push( [ 'generateblocks/container', {} ] );
+	}
+	return innerBlocks;
+}
+
+function addTemplateSelectors( context, props ) {
+	const layouts = {
+		'': 1,
+		'repeat(2, 1fr)': 2,
```
iansvo, line 35: I'd recommend you add the minmax pattern here for the repeats to avoid any grid blowouts (which for some reason are still a thing). 

Ex: `repeat(2, minmax(0px, 1fr))`

src/blocks/container/templates.js
```diff
@@ -0,0 +1,198 @@
...
+	const repeatCount = layouts[ layoutKey ] || 1;
+	const innerBlocks = [];
+	for ( let i = 0; i < repeatCount; i++ ) {
+		innerBlocks.push( [ 'generateblocks/container', {} ] );
+	}
+	return innerBlocks;
+}
+
+function addTemplateSelectors( context, props ) {
+	const layouts = {
```
iansvo, line 33: What do you think about including an auto-fill and auto-fit option here?

src/blocks/container/templates.js
```diff
@@ -0,0 +1,198 @@
...
+				attributes: {
+					display: 'grid',
+					gridTemplateColumns: '3fr 1fr',
+				},
+			},
+			{
+				id: 'three-column-1-1-2',
+				label: __( 'Three Columns (1/1/2)', 'generateblocks' ),
+				labelAsTooltip: true,
+				icon: buildLayout( layouts, '1fr 1fr 2fr' ),
```
iansvo, line 122: Seems like there is a lot of repetition here with the grid-template-columns. Maybe you can set the layouts up as a key so you can do `layouts['three-column-1-1-2']` to get the value `1fr 1fr 2fr` and so on.

src/extend/inspector-control/controls/layout/attributes.js
```diff
@@ -4,48 +4,49 @@ export default function getLayoutAttributes( defaults ) {
...
 		'flexWrap',
+		'flexBasis',
 		'alignItems',
 		'justifyContent',
 		'columnGap',
 		'rowGap',
 		'position',
 		'overflowX',
 		'overflowY',
+		'gridTemplateColumns',
```
iansvo, line 15: I think we should have align|justify-self in the mix, which is pretty useful in Grid implementations.

src/blocks/container/components/GridBlockSettings.js
tomusborne: Yea, we certainly could. This code was mainly copied from Global Styles which doesn't have an icon function/library.

src/extend/inspector-control/controls/layout/components/GridColumnSelector.js
```diff
@@ -0,0 +1,44 @@
+import classnames from 'classnames';
+import { Button } from '@wordpress/components';
+
+export function GridColumnSelector( { onClick, value } ) {
+	const layouts = {
```
iansvo, line 5: If this is in 2 places maybe it can be an export and shared by both?

src/extend/inspector-control/controls/layout/index.js
```diff
@@ -213,6 +289,67 @@ export default function Layout( { attributes, setAttributes, computedStyles } )
...
+								<TextControl
+									id="grid-template-columns"
+									value={ deviceAttributes.gridTemplateColumns }
+									onChange={ ( value ) => setDeviceAttributes( { gridTemplateColumns: value } ) }
+								/>
+								<Button
+									size="small"
+									onClick={ () => setShowGridTemplateColumns( ! showGridTemplateColumns ) }
+									icon={ () => <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 256"><rect width="256" height="256" fill="none" /><rect x="32" y="56" width="192" height="144" rx="8" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="96" y1="56" x2="96" y2="200" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="160" y1="56" x2="160" y2="200" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="32" y1="104" x2="224" y2="104" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /><line x1="32" y1="152" x2="224" y2="152" fill="none" stroke="currentColor" strokeLinecap="round" strokeLinejoin="round" strokeWidth="12" /></svg> }
```
iansvo, line 323: Can we use getIcons here?

src/blocks/container/templates.js
tomusborne: Yea not sure why that's there - does `aria-label` make sense here?

iansvo (commented):
Overall looking solid, mostly some nit things. I haven't checked it in browser yet though.

tomusborne:
@iansvo Made some updates based on your feedback - thanks!

`auto-fill` and `auto-fit` - any suggestions of how to fit these into our presets? Do we have to define a min width for our users?

`align-self` and `justify-self` would need to be added to our existing flex/grid alignment options. Maybe a different PR?

PR #1231, opened 2024-04-25
Tweak: Use ColorPicker component

tomusborne:
This PR makes use of our `@edge22/components` package for the `ColorPicker` component.


includes/general.php
```diff
@@ -176,6 +176,14 @@ function generateblocks_do_block_editor_assets() {
 		array( 'wp-components' ),
 		filemtime( GENERATEBLOCKS_DIR . 'dist/editor-sidebar.css' )
 	);
+
+	$component_asset_info = generateblocks_get_enqueue_assets( 'components' );
+	wp_register_style(
+		'generateblocks-components',
+		GENERATEBLOCKS_DIR_URL . 'dist/components.css',
```
iansvo, line 183: It might be possible that we load the same thing twice. Maybe we should do some kind of check to see if the components.css file was enqueued already before loading another one?

iansvo (commented):
Do we still need to update the ColorPicker component to include that edge case for the older attribute value?

tomusborne:
I already did that in the latest components package.

As for the enqueue, how does this work if we use the same handle in both places? I have a feeling it will only load once since the handle is the same. Which one it loads is a different question, though..

iansvo:
> I already did that in the latest components package.

I was looking at the repo and I'm not seeing a commit: 

image

Did you perhaps publish and not commit the changes?

> As for the enqueue, how does this work if we use the same handle in both places? I have a feeling it will only load once since the handle is the same. Which one it loads is a different question, though..

I feel like if the handle is the same it won't do it twice, but I suppose we'll just need to double check. I _believe_ if the script is registered already then re-registering it will overwrite the previous registration though. It might mean that we need to lock the version to be the latest so it always pulls the same version, but we can solve for that easily enough I think.

tomusborne:
This is it here: https://github.com/EDGE22-Studios/components/commit/ec5f9ea52436391d3541a59e51e038d46246f60e

> It might mean that we need to lock the version to be the latest so it always pulls the same version, but we can solve for that easily enough I think.

This would be smart.

tomusborne:
@iansvo We good to go on this for now?

PR #1245, opened 2024-06-07
Fix text-decoration for headings

iansvo:
Removes extra style that caused the rich-text span to not display inline styles correctly


iansvo:
@tomusborne Updated to release 1.10.0, let me know if this should be 2.0 instead.

tomusborne:
@iansvo Yea let's put this in 2.0.

tomusborne (approved):

PR #1265, opened 2024-08-05
Bug: Missing script deps for packages

iansvo:
The new package entries were missing the proper deps list. This retrieves the deps from the package itself to ensure they’re correct.

### DIFF
--- a/includes/general.php
+++ b/includes/general.php
@@ -213,7 +213,14 @@ function generateblocks_do_block_editor_assets() {
 		);
 
 		foreach ( $edge22_packages as $name => $version ) {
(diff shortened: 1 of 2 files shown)


includes/general.php
```diff
@@ -234,16 +241,6 @@ function( $package_name ) {
 		}
 	}
 
-	$styles_builder_assets = generateblocks_get_enqueue_assets( 'styles-builder' );
```
iansvo, line 237: This was an extra enqueue from a test that I forgot to remove previously.

tomusborne (approved):

PR #1266, opened 2024-08-06
Tweak - Adjust legacy blocks check to use version

iansvo:
Simple change to use the free plugin’s version to determine if legacy blocks should be shown in the inserter. The previously added hook remains in place.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -2002,10 +2002,8 @@ function generateblocks_str_starts_with( $string, $prefix ) {
 function generateblocks_should_show_legacy_blocks() {
 	$show_legacy_blocks = false;
 
-	if ( defined( 'GENERATEBLOCKS_PRO_VERSION' ) ) {
-		if ( ! class_exists( 'GenerateBlocks_Block_Accordion_Item' ) ) {
-			$show_legacy_blocks = true;
-		}
+	if ( defined( 'GENERATEBLOCKS_VERSION' ) && GENERATEBLOCKS_VERSION < 2.0 ) {
+		$show_legacy_blocks = true;
 	}
 
 	return apply_filters(


includes/functions.php
```diff
@@ -2002,10 +2002,8 @@ function generateblocks_str_starts_with( $string, $prefix ) {
 function generateblocks_should_show_legacy_blocks() {
 	$show_legacy_blocks = false;
 
-	if ( defined( 'GENERATEBLOCKS_PRO_VERSION' ) ) {
-		if ( ! class_exists( 'GenerateBlocks_Block_Accordion_Item' ) ) {
-			$show_legacy_blocks = true;
-		}
+	if ( defined( 'GENERATEBLOCKS_VERSION' ) && GENERATEBLOCKS_VERSION < 2.0 ) {
```
tomusborne, line 2005: @iansvo We need to check the `GENERATEBLOCKS_PRO_VERSION` here.

The check above has the plugin checking itself, so the check will never be true.

I believe GB 2.0 will be released along with GBP 1.8.0.

iansvo: Couldn't someone have GBP 1.8 and then use a version of GB that's < 2.0? What is the significance of checking the pro plugin's version here? I see what you mean regarding this comparison, maybe we just need to have a filter and that's it? Once the plugin is updated it will just start hiding those blocks automatically and then we can use a filter or something other UI setting (I think that actually makes sense here). 

We could even piggyback on the "Allowed Blocks" setting which I believe would do the same thing (leave the blocks in content just take them out of the inserter).

tomusborne: This code only runs if the user is using GB 2.0 or greater.

The purpose of this check is to prevent the user from using the 2.0 blocks (in this release) if they're still relying on the blocks in GBP (prior to 2.0). For example, the Accordion block in GBP < 1.8 depends on the "old" blocks in GB < 2.0.

So, new blocks in 2.0 should be available if:

1. The user does not have GBP.
2. The user is using GBP > 1.8.0.

The old blocks should remain available if:

1. The user is using GBP and the version is < 1.8.0

iansvo: Roger, I'll update

iansvo: Ok I updated this to use version_compare so it correctly works with the -alpha, -beta, and -rc version strings.

PR #1269, opened 2024-08-06
Fix: Legacy blocks showing

tomusborne:
Legacy blocks are showing even when using GBP 1.8.0.


iansvo (approved):
It's weird because an alpha is considered less than the version, good fix.

PR #1271, opened 2024-08-09
Bug - Resolve runtime errors in editor

iansvo:
This fixes a number of existing runtime errors and removes wp-polyfill from the webpack output.


tomusborne:
@iansvo I've seen these errors so many times, but now that I'm trying to test this I'm not getting them anymore lol

iansvo:
In retrospect I probably should've added some description/notes to this PR to clarify, but this happens only when a Headline block is using dynamic data. 

If the InnerContent component is anything other than a RichText component (the default for a headline) then it was erroring because it was adding props that only the RichText accepts. I also just removed an extra option from the webpack config that I don't think we actually needed in the first place.

PR #1279, opened 2024-08-21
Feature: Add `linkTo` feature to dynamic tags

tomusborne:
Blocked by #1278 

This PR adds a `linkTo` param to the dynamic tags. This allows you to choose whether to automatically link the dynamic text or not.

It also runs all dynamic tag output through a single filter, so users can filter their output based on the tag/options.

### DIFF
--- a/includes/dynamic-tags/class-dynamic-tag-callbacks.php
+++ b/includes/dynamic-tags/class-dynamic-tag-callbacks.php
@@ -41,7 +41,7 @@ class GenerateBlocks_Dynamic_Tag_Callbacks extends GenerateBlocks_Singleton {
 		if ( $link ) {
 			$output = sprintf(
 				'<a href="%s">%s</a>',
-				$link,
+				esc_url( $link ),
 				$output
 			);
 		}
--- a/includes/dynamic-tags/class-dynamic-tag-callbacks.php
+++ b/includes/dynamic-tags/class-dynamic-tag-callbacks.php
@@ -21,7 +21,7 @@ class GenerateBlocks_Dynamic_Tag_Callbacks extends GenerateBlocks_Singleton {
 	 * @param string $output The output.
 	 * @param array  $options The options.
 	 */
-	private static function withLink( $output, $options ) {
+	private static function with_link( $output, $options ) {
 		if ( empty( $options['linkTo'] ) ) {
 			return $output;
 		}
@@ -56,7 +56,7 @@ class GenerateBlocks_Dynamic_Tag_Callbacks extends GenerateBlocks_Singleton {
 	 * @param array  $options The options.
 	 */
 	private static function output( $output, $options ) {
-		$output = self::withLink( $output, $options );
+		$output = self::with_link( $output, $options );
 		$output = apply_filters( 'generateblocks_dynamic_tag_output', $output, $options );
 
 		return $output;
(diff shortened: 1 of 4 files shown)


includes/dynamic-tags/class-dynamic-tag-callbacks.php
```diff
@@ -15,16 +15,64 @@
...
+	 * @param string $output The output.
+	 * @param array  $options The options.
+	 */
+	private static function withLink( $output, $options ) {
+		if ( empty( $options['linkTo'] ) ) {
+			return $output;
+		}
+
+		$id      = GenerateBlocks_Dynamic_Tags::get_id( $options );
+		$link_to = $options['linkTo'];
```
iansvo, line 30: Need sanitization/escaping on this value

includes/dynamic-tags/class-dynamic-tag-callbacks.php
```diff
@@ -15,16 +15,64 @@
  * @since 2.0.0
  */
 class GenerateBlocks_Dynamic_Tag_Callbacks extends GenerateBlocks_Singleton {
+	/**
+	 * Wrap a link around the output.
+	 *
+	 * @param string $output The output.
+	 * @param array  $options The options.
+	 */
+	private static function withLink( $output, $options ) {
```
iansvo, line 24: How dare you

src/dynamic-tags/components/DynamicTagSelect.jsx
```diff
@@ -237,13 +248,26 @@ export function DynamicTagSelect( { onInsert, tagName, value: selectedValue } )
...
+					) }
+
 					<TextControl
 						label={ __( 'Dynamic tag to insert', 'generateblocks' ) }
 						value={ dynamicTagToInsert }
 						onChange={ ( value ) => setDynamicTagToInsert( value ) }
 					/>
 
-					{ !! hasSelection && ! interactiveTagNames.includes( tagName ) && (
+					{ !! hasSelection && ! interactiveTagNames.includes( tagName ) && ! linkTo && (
```
iansvo, line 270: These types of lengthy conditionals can get a bit verbose for render JSX. We should change this to a variable to make it super clear (and also easier to filter if we ever want to do that later).

src/dynamic-tags/components/DynamicTagSelect.jsx
```diff
@@ -149,6 +154,10 @@ export function DynamicTagSelect( { onInsert, tagName, value: selectedValue } )
 			tags.push( `multiple=${ commentsCountText.multiple }` );
 		}
 
+		if ( linkTo ) {
+			tags.push( `linkTo=${ linkTo }` );
```
iansvo, line 158: The variable name "tags" here seems confusing. It seems like it should be named "options" if anything (since that's what we call those elsewhere).

includes/dynamic-tags/class-dynamic-tag-callbacks.php
```diff
@@ -68,20 +121,23 @@ public static function get_modified_date( $options ) {
...
 
 		if ( ! $image ) {
-			return '';
+			return self::output( $output, $options );
 		}
 
-		return $image[0];
+		$output = $image[0];
+
+		return self::output( $output, $options );
```
iansvo, line 140: Can we add support for an image size option here? This should support string or int array just like the called function does.

iansvo (changes requested):
Only one requested change, rest are largely suggestions/feedback.

includes/dynamic-tags/class-dynamic-tag-callbacks.php
tomusborne: This value isn't output anywhere, but I escaped the actual link down below.

includes/dynamic-tags/class-dynamic-tag-callbacks.php
tomusborne: Let's improve this in a separate PR - the image stuff needs some playing with so we get it right.

PR #1288, opened 2024-09-06
Feature: Inline background image

tomusborne:
This adds an inline background image option to our Container block.


src/blocks/element/components/InlineBackgroundImage.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+		if ( ! inlineBackgroundPart ) {
+			return '';
+		}
+
+		return inlineBackgroundPart.split( 'url(' )[ 1 ].split( ')' )[ 0 ];
+	}, [ htmlAttributes?.style ] );
+
+	function onChange( value ) {
+		const style = htmlAttributes?.style
```
iansvo, line 29: Seems like you could use destructuring here to do the same thing but a bit more readable?

```
const { style = '' } = htmlAttributes;
```

src/blocks/element/utils.js
```diff
@@ -1,10 +1,22 @@
 export const convertInlineStyleStringToObject = ( styleString ) => {
```
iansvo, line 1: One thought here....if we are talking about inline styles and we have a selected block (via useSelectedBlockElement) then we could just get the styles attribute off the element directly and not have to build the object at all.

iansvo (commented):
Couple of feedback items, no hard required chagnes though.

src/blocks/element/utils.js
tomusborne: Would be simpler, but we need this to work when the block isn't selected as well.

I tried to do a `ref.current.style`, but that gave me a ton of things we don't need.

We can improve this in a future PR if we have a better way.

PR #1291, opened 2024-09-10
Refactor: Dynamic tags

tomusborne:
This PR changes how dynamic tags are formatted.

Before:
`{post_title postId=123|linkTo=post|renderIfEmpty=true}`

Now:
`{post_title id:123|link:post|renderIfEmpty}`

Main differences:

1. We now use a `:` to separate keys and values.
2. We shortened the keys where possible.
3. Keys can have no value when `true` is assumed.
4. Tags that are not registered will not try to replace themselves in the editor.


iansvo:
I like it! Looks solid and I think this should make things easier to use, for sure.

iansvo (approved):

PR #1298, opened 2024-09-14
Fix: Target and rel bug

tomusborne:
When you add `target="_blank"` to any link in Gutenberg, it will automatically add `rel="noopener"` to the tag.

This was causing an invalid block error when you choose to open a new tab on your `a` blocks in v2 blocks, as the save component wasn't matching the edit component.

This PR ensures that we add `rel="noopener"` if `target="_blank"` is present.


iansvo (approved):
Looks good, questionable other things in the file aside :)

PR #1300, opened 2024-09-16
Feature: Add onboarders for both styles and dynamic tags

tomusborne:
This PR adds block setting panels showing users where to find the new Styles panel and the Dynamic Tags options.

This should help people who are used to finding both of these things in their block settings.

Clicking the "Got it" button will save the preference to user meta and no longer show the panel.

dynamic-data-onboarder
styles-onboarder


src/components/dynamic-tags-onboarder/DynamicTagsOnboarder.jsx
```diff
@@ -0,0 +1,63 @@
...
+		borderRadius: '2px',
+	};
+
+	return (
+		<OpenPanel
+			title={ __( 'Dynamic Data', 'generateblocks' ) }
+			panelId="dynamic-content"
+		>
+			<p style={ { marginBottom: '0' } }>{ __( 'Dynamic data options can be found by clicking on the dynamic tags icon in your block toolbar and/or block settings.', 'generateblocks' ) }</p>
+			<img
```
iansvo, line 38: Can we add the image dimensions here to avoid layout shift?

src/components/dynamic-tags-onboarder/DynamicTagsOnboarder.jsx
```diff
@@ -0,0 +1,63 @@
...
+		) {
+			setShowOnboard( true );
+		}
+	}, [ user?.id, JSON.stringify( user?.meta ) ] );
+
+	if ( ! screenshot || ! showOnboard ) {
+		return null;
+	}
+
+	const style = {
```
iansvo, line 27: Seems like we should make these a reusable component or something instead of having the inline styles hardcoded in 2 places.

iansvo (commented):
I feel like the onboarder itself should be a reusable component since it's essentially identical in both spots, etc.

tomusborne:
Updated this PR with changes we discussed today.

dynamic-data-notice
styles-notice

PR #1302, opened 2024-09-16
Add post excerpt, minor bug fix for filters

iansvo:
Fixes order of trunc/replace


tomusborne:
Removed the `{post_excerpt}` tag for now. We need to think on this a bit more, but the other fixes in here are good.

PR #1309, opened 2024-09-17
Tweak: HTML attributes handling

tomusborne:
This PR puts the logic for our HTML attributes into one HOC.

It also will automatically remove any attributes that have no value, unless that attribute is allowed to have no value.


includes/functions.php
```diff
@@ -2100,3 +2100,81 @@ function generateblocks_with_html_attributes( $html_attributes, $block_attribute
...
+
+/**
+ * Get the HTML attributes for a block.
+ * This function is only used if generateblocks_get_processed_html_attributes() returns nothing.
+ *
+ * @since 2.0.0
+ * @param string $block_slug The block slug.
+ * @param array  $attributes The block attributes.
+ */
+function generateblocks_get_backup_html_attributes( $block_slug, $attributes ) {
```
iansvo, line 2145: Would it make sense to just include this in the processed attributes function if the return is empty? Otherwise you have to remember that's there and it seems like it would always be called if the original is empty, etc.

iansvo (approved):
Seems solid overall, no major concerns. I like the streamlined attribute handling for sure.

PR #1319, opened 2024-09-24
Tweak: Try dynamic tag preview mode

tomusborne:
In an effort to minimize confusion and prevent editing dynamic tags by mistake, this PR adds a "Preview" and "Edit" mode to the Text block when a dynamic tag is present.

https://github.com/user-attachments/assets/81ae6fb5-e5da-4168-9ce8-59e0f4ea2ffa


iansvo:
@tomusborne This looks pretty fantastic. Once I wrap up this other PR, I think it might make sense to just tackle this as a separate merge so it doesn't get tied into the other work I was doing (which is already a big changeset).

PR #1340, opened 2024-09-27
Tweak: Unify block appender across blocks

tomusborne:
This makes it so our new blocks that have inner blocks all use the same block appender.


src/blocks/query/edit.js
```diff
@@ -58,6 +59,7 @@ function EditBlock( props ) {
 				'generateblocks/query-page-numbers',
 				'generateblocks/element',
 			],
+			renderAppender: () => <BlockAppender clientId={ clientId } isSelected={ isSelected } attributes={ attributes } />,
```
iansvo, line 62: Does this need to have a function wrapper if it's returning a JSX element?

iansvo (commented):
Made a few notes on here, overall nice to see things getting more streamlined!

src/blocks/query/edit.js
tomusborne: In my tests, yes. Otherwise the block errors out.

PR #1375, opened 2024-10-12
Tweak: Switch required tag option logic

tomusborne:
This PR flips the logic when it comes to dynamic tags that have no value.

By default, all tags are required to have a value. If one of the tags in a block has no value, the entire block will not render.

This can be prevented by adding `required:false` to any non-essential tags.

### DIFF
--- a/src/dynamic-tags/components/DynamicTagSelect.jsx
+++ b/src/dynamic-tags/components/DynamicTagSelect.jsx
@@ -164,7 +164,7 @@ export function DynamicTagSelect( { onInsert, tagName, selectedText, currentPost
 		multiple: __( '%s comments', 'generateblocks' ),
 	} );
 	const [ linkTo, setLinkTo ] = useState( '' );
-	const [ renderIfEmpty, setRenderIfEmpty ] = useState( false );
+	const [ required, setRequired ] = useState( true );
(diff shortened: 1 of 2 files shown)


src/dynamic-tags/components/DynamicTagSelect.jsx
```diff
@@ -164,7 +164,7 @@ export function DynamicTagSelect( { onInsert, tagName, selectedText, currentPost
 		multiple: __( '%s comments', 'generateblocks' ),
 	} );
 	const [ linkTo, setLinkTo ] = useState( '' );
-	const [ renderIfEmpty, setRenderIfEmpty ] = useState( false );
+	const [ required, setRequired ] = useState( true );
```
iansvo, line 167: I am thinking we should include a configuration option that forces the tag to be required. For instance, on registration you could say indicate that the tag is _always_ required. I feel like there could be scenarios where that's useful, but maybe I'm over thinking it.

tomusborne: That would be cool. Here we always default to required, so the user has to opt-out.

Are there any tags in our current list that would not be required by default? I feel like it really depends on the context of the block.

PR #1379, opened 2024-10-15
Fix - Query Loop broken when queryData is empty

### LINKED ISSUE #722: Query loop block: by product categories, properties, filtering
I would like to have a function that allows me to filter one (or more) product category (WooCommerce) with the query loop. Currently I can display all or individual products, but not filter by categories, properties ect. This would be really great if this works.

Practical example: A page with a query loop that gives me 6 products from category XYZ.

iansvo:
Fixes https://github.com/tomusborne/generateblocks-pro/issues/722


iansvo:
@tomusborne Added a few more commits/fixes to this. Should be pretty tightened up now. The last thing that remains is the question of what we're doing with Elements. I'm not exactly sure what the ideal UX is and I'm also not sold on the idea of just throwing every field group and key in there either. For one thing, we'd have to query the type for each CPT, or else add logic that checks the element locations to determine what post type it shows up for and then attempt to whittle it down, but overall it just feels like a lot of logic that could easily be done wrong if we're not careful.

That's not to say I'm fundamentally against the idea, I just want it to be a well considered and thought out UX so we don't spend undue time working on it. Maybe this is a good question for Jonathan and Kyle et al to give us feedback on?

PR #1395, opened 2024-10-24
Dynamic Tags - Update image handling, fix various small issues

iansvo:
### New
- Added author_avatar_url tag with options
- Added new Link options: Post Meta, Author Archive, Author Email (with mailto), and Author Meta
- Added new 'image-sizes' dynamic tag support, which will enable the image size dropdown and the `size` option for that tag.
- Added new 'instant-pagination' dynamic tag support for tags that act like a pagination link inside a Query. This automatically updates the markup before tag replacement to support instant pagination.

### Improved
- Cleaned up all tag-specific logic in the DynamicTagSelect and updated registration calls to use the new declarative options key.
- API Version of _most_ blocks was updated to v3. Button-container was not updated so we don't force the editor iframe yet. This will be updated again once the requisite GP updates are released. 
- Various hooks that had been moved to edge22/components were removed from the plugin code.

### Fixed
- Fixed issue that caused the Image (Media) block to reload when the block was selected/unselected.
- Fixed some small UI and state things
- Updated styles builder to fix search/filtering visibility logic


src/blocks/media/edit.js
```diff
@@ -54,9 +54,11 @@ function EditBlock( props ) {
 	}, [ tagName ] );
 
 	const blockProps = useBlockProps();
+	const shouldWrapBlock = isSelected || ( 'img' === tagName && ! temporaryURL && ! htmlAttributes?.src );
 	const elementAttributes = {
 		className: classNames.join( ' ' ).trim(),
 		'data-block': clientId,
+		'data-should-wrap': shouldWrapBlock,
```
tomusborne, line 61: Do we still need this?

src/blocks/media/editor.scss
tomusborne, line 1: Is this file doing anything?

includes/dynamic-tags/class-dynamic-tags.php
```diff
@@ -287,6 +341,14 @@ public function register() {
 				'type'        => 'term',
 				'supports'    => [ 'link', 'source' ],
 				'description' => __( 'Get a list of terms for the specified post.', 'generateblocks' ),
+				'options'     => [
+					'sep' => [
+						'type'        => 'text',
+						'label'       => __( 'Separator', 'generateblocks' ),
+						'placeholder' => ', ',
```
tomusborne, line 348: I don't think this can/should be a default. What do you I do if I want no separator?

src/hooks/useTaxonomies.js
```diff
@@ -1,10 +1,22 @@
...
-);
+		const args = { per_page: -1 };
+
+		/**
+		 * Certain post types are excluded from this behavior since they are "pattern-like" and might display in a different
+		 * content type than the one being edited.
+		 */
+		const excludedPostTypes = [ 'gp_elements', 'wp_block' ];
+
+		if ( postType && ! excludedPostTypes.includes( postType ) ) {
```
tomusborne, line 16: The taxonomies dropdown is empty for me, even if I specify a regular `post` as my source.

includes/dynamic-tags/class-dynamic-tags.php
iansvo: Only thing I can think of is just doing like sep:false or something like that. But that's not exactly obvious. Maybe we can include a checkbox value for the sep and if it's set, then we'd use the sep value?

src/blocks/media/editor.scss
iansvo: No it was there before though, just with the wrong name. We can remove it if we want, but we should do the same with the elements block's editor.scss too.

src/blocks/media/edit.js
iansvo: No I think we can just remove that.

includes/dynamic-tags/class-dynamic-tags.php
tomusborne: Is it bad just to have the sep in there by default?

tomusborne (approved):

PR #1396, opened 2024-10-24
Dynamic Tags - Bugfixes and tweaks

iansvo:
- Fixed issue that caused Autocompletes to not retain a custom value. - Implemented Autocomplete for dynamic tags and added groups. - Added sorting logic to show the Post group first when in a WP_Query.


tomusborne:
Pushed a couple of bug fixes.

Seems that using the `Autocomplete` component for selecting the dynamic tag is causing some breakages.

For example, you can see here is causing some re-renders, so we're losing our parsing:

https://github.com/user-attachments/assets/94484996-f6c0-4e54-83ad-d5f7f82987f0

Not sure what could be up there?

tomusborne (approved):

PR #1402, opened 2024-10-28
Dynamic Tags: Add post_excerpt

iansvo:
Adds post excerpt along with a few minor tweaks


iansvo:
@tomusborne This is updated now, I redid the logic to properly support default values this time also.

PR #1403, opened 2024-10-29
Tweak: Dashboard

tomusborne:
This changes how our Dashboard is built, and simplifies it for now until we decide what should go in there.


iansvo:
@tomusborne My first reaction looking at this was: Why isn't this all on one page?

Most of the pages don't have a ton on them. The local patterns one is the only outlier so that one feels sensible as it's own page. But there's so little on every other one, it seems much more efficient to just load them all up at once.

tomusborne:
Was playing with merging the Dashboard and Settings pages last night. For sure possible, but more involved than it should be. I can take another stab at it.

tomusborne:
Removed the "Dashboard" page for now as it didn't offer a ton of value.

Going to merge this so we can get #1416 merged.

PR #1406, opened 2024-10-29
Fix: Container toolbar appenders

tomusborne:
This PR does a couple of things.

1. Reverts the version 1 block toolbar appenders to how they were, so they will continue to work as they always have.
2. Moves the version 2 block toolbar appenders into the proper file.
3. Conditionally shows the correct buttons based on the block version you have set.


src/editor/container-toolbar-appenders.js
```diff
@@ -0,0 +1,142 @@
+import { ToolbarGroup, ToolbarButton } from '@wordpress/components';
+import { addFilter } from '@wordpress/hooks';
+import { __ } from '@wordpress/i18n';
+import { useSelect, useDispatch } from '@wordpress/data';
+import { createBlock } from '@wordpress/blocks';
+import { createHigherOrderComponent } from '@wordpress/compose';
+import { BlockControls } from '@wordpress/block-editor';
+import useInnerBlocksCount from '@hooks/useInnerBlocksCount';
```
iansvo, line 8: We really need to make a lint rule for import order (local imports come last, stylesheet or nonjs local imports come at the end of those, etc).

src/extend/block-controls/toolbar-appenders/index.js
```diff
@@ -55,21 +55,17 @@ const withToolbarAppenders = createHigherOrderComponent( ( BlockEdit ) => {
 			const containerAttributes = {};
 
 			if ( 'row' === layout ) {
-				containerAttributes.styles = {
-					display: 'flex',
-					flexDirection: 'row',
-				};
+				containerAttributes.display = 'flex';
```
iansvo, line 58: This logic is all for the v1 stuff, right?

iansvo (approved):
LGTM, assuming I didn't misunderstand the question I asked in one of my comments.

src/extend/block-controls/toolbar-appenders/index.js
tomusborne: Yea, the entire `/extend/` folder should ideally be v1 only, making it easier for us to dequeue `blocks.js` in the future.

PR #1414, opened 2024-10-30
Dynamic Tag - Update wrapper characters

iansvo:
Now dynamic tags use a double curly brace pattern.


tomusborne (approved):
Seems to all work!

PR #1431, opened 2024-11-04
Refacto: Remove unused function

tomusborne:
This removes a function that ended up never being used, as well as a variable.


iansvo (approved):
LGTM

Issue #1446, opened 2024-11-06
Styles do no apply to multiple selected blocks and only the last one gets styled


Kovas:
### Description
When I select multiple blocks and change their styles, they do not appear correctly on the frontend.

### Steps to reproduce
1.  Select multiple containers.
2.  Change the background color.

#### Expected behavior
All containers have the same background color.

#### Actual behavior
Containers change color in the editor, but only the last one gets styles applied on the frontend.

tomusborne:
Thanks for reporting this! Should be fixed in the next release.

PR #1450, opened 2024-11-06
WIP: Multi-selected block styles

tomusborne:
closes https://github.com/tomusborne/generateblocks/issues/1446
requires https://github.com/EDGE22-Studios/block-styles/pull/8

This is an attempt at fixing the above issue.

It still doesn't fix the fact that changing one single property when multi-selecting blocks changes the entire styles object for all blocks. Not sure how feasible that is at this point, to be honest. Worth a discussion.


src/hoc/withSetBlockAttributes.js
```diff
@@ -0,0 +1,93 @@
...
+		uniqueId,
+	} = attributes;
+
+	const multiSelectedBlocks = useSelect( ( select ) => (
+		select( 'core/block-editor' )?.getMultiSelectedBlocks() || []
+	), [] );
+
+	const { updateBlockAttributes } = useDispatch( 'core/block-editor' );
+
+	async function setBlockAttributes( attrs ) {
```
iansvo, line 31: I don't fundamentally dislike this approach. One thing I'd note though is this function almost definitely needs a useCallback wrapper so it doesn't regenerate itself each time. It has a lot of logic and reactive pieces, etc. Could help reduce renders and improve perf if we end up using this.

tomusborne:
Handled by https://github.com/tomusborne/generateblocks/pull/1485

PR #1460, opened 2024-11-07
Query Block - Use real WP_Query args

iansvo:
Related: https://github.com/tomusborne/generateblocks-pro/pull/755

This PR updates our query block parameters to use the real ones from a WP_Query and not the ones for the REST API/getEntityRecord. This ensures data is correct on the frontend and backend. 

Some issues to address before merge.

- [x] `SelectPosts` component isn't searching
- [x] The REST endpoints likely shouldn't be within the dynamic tag classes.
- [x] v1 Query Loop block can't add new query params


tomusborne:
Great work on this @iansvo! It feels faster than before as well somehow.

Couple things need fixing (Query related, not so much this PR related):

1. Missing the Exclude current post (and other GBP integrations). You can find the current integrations with v1 here: https://github.com/tomusborne/generateblocks-pro/blob/master/src/extend/query-loop/index.js
2. We should make the Post Type option an array so the user can choose more than one post type.

I'll keep testing for bugs. We may want to address the above points in a new PR anyways.

tomusborne:
Some more issues here.

1. The `SelectPost` search isn't working.
2. The REST endpoints likely shouldn't be within the dynamic tag classes.

tomusborne (approved):

PR #1472, opened 2024-11-11
Fix: Block multi-selection CSS generation

### LINKED ISSUE #1446: Styles do no apply to multiple selected blocks and only the last one gets styled

tomusborne:
closes #1446 
alternative to #1450 
requires https://github.com/EDGE22-Studios/block-styles/pull/7

Major benefit here is it doesn't change how CSS is generated (via a hook). Therefore it doesn't break other functionality like #1450 (copy/paste styles).

### DIFF
--- a/src/hoc/withStyles.js
+++ b/src/hoc/withStyles.js
@@ -115,6 +115,7 @@ export function withStyles( WrappedComponent ) {
 
 		useGenerateCSSEffect( {
 			selector,
+			getSelector,
 			styles: frontendStyles,
 			setAttributes,
 			getCss,


tomusborne:
Handled by https://github.com/tomusborne/generateblocks/pull/1485

PR #1475, opened 2024-11-11
Feature: Add pagination type option

tomusborne:
This changes from our "Instant pagination" option to a "Pagination type" option. This will allow us to add more types in the future.


includes/dynamic-tags/class-dynamic-tags.php
```diff
@@ -711,7 +711,8 @@ function( $key ) {
 	 */
 	public function before_tag_replace( $content, $args ) {
 		if ( in_array( 'instant-pagination', $args['supports'], true ) ) {
-			$instant_pagination = $args['instance']->context['generateblocks/instantPagination'] ?? true;
+			$pagination_type    = $args['instance']->context['generateblocks/paginationType'] ?? '';
+			$instant_pagination = 'instant' === $pagination_type;
```
iansvo, line 715: let's update the pagination type to be a constant.

iansvo (changes requested):
Only change requested is to use static consts for the values like `instant` and such.

PR #1483, opened 2024-11-14
New Multiselect Component

iansvo:
This PR adds support for the new MultiSelect component. Currently this is applied to the Include Posts and Post Type params, but we should use it for anything that requires multiple selection.

https://github.com/EDGE22-Studios/components/pull/6

Note: This PR needs to be updated with the newest components version in the above PR before merging.


tomusborne:
@iansvo This is working great.

One thing I've noticed while testing (don't think it's a part of this PR):


https://github.com/user-attachments/assets/09d93881-1cd3-4c66-841d-1ff7cf727faf

iansvo:
@tomusborne This is kind of expected at this point because I didn't update that specific component to use the new pattern. 

I think at this point I need to just swap those other things out so we can test if it all works. The actual multiselection part in general works as expected based on what we're both seeing.

tomusborne:
@iansvo So even though the control is using the new component, it's not actually saving the values so that they apply to the query?

If that's the case, then yea, the component itself seems to be working as intended. Let's swap out the other things so we can make sure there's no issues before merging.

tomusborne (approved):

PR #1485, opened 2024-11-15
Refactor: Improve how local block styles are set

tomusborne:
This PR makes it so our local blocks use the `styles` attribute as their styles store, instead of having a separate styles store we have to constantly sync with.

requires https://github.com/EDGE22-Studios/block-styles/pull/9
requires https://github.com/EDGE22-Studios/styles-builder/pull/17


iansvo (approved):
After a few tweaks made in the other PRs, this is working as expected in local testing.

PR #1488, opened 2024-11-15
Fix meta handler id handling

iansvo:
Fixes comparison in get_meta that caused lookups to fail if a numeric ID was passed as a string.

### DIFF
--- a/includes/class-meta-handler.php
+++ b/includes/class-meta-handler.php
@@ -205,7 +205,7 @@ class GenerateBlocks_Meta_Handler extends GenerateBlocks_Singleton {
 			$callable
 		);
 
-		if ( is_int( $id ) ) {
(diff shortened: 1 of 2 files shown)


includes/class-meta-handler.php
```diff
@@ -205,7 +205,7 @@ public static function get_meta( $id, $key, $single_only = true, $callable = nul
 			$callable
 		);
 
-		if ( is_int( $id ) ) {
+		if ( is_int( $id ) || is_string( $id ) ) {
```
tomusborne, line 208: Don't think this will work. Entire purpose of the check is to see if it's a number.

Let's just use `is_numeric()` instead.

iansvo: Yeah sure that's fine
```suggestion
		if ( is_numeric( $id ) ) {
```

PR #1513, opened 2024-11-20
Fix dynamic bg images

### LINKED ISSUE #1357: GB 2.0 Pre-alpha - Container Dynamic Background does not render in the editor
### Description
Adding a dynamic data tag to the background of a container block does not render in the editor.

iansvo:
closes #1357


src/hoc/withHtmlAttributes.js
```diff
@@ -68,16 +69,46 @@ export function withHtmlAttributes( WrappedComponent ) {
...
 		const { style = '', href, ...otherAttributes } = htmlAttributes;
-		const inlineStyleObject = typeof style === 'string'
-			? convertInlineStyleStringToObject( style )
-			: style;
+
+		useEffect( () => {
+			async function getReplacements() {
+			// Check if any replacements need to be made if not, do nothing.
+				if ( ! style.includes( '{{' ) ) {
+					return style;
```
tomusborne, line 79: @iansvo If we return early and never `setWithReplacements`, then the `inlineStyleObject` below would never have anything in it, right? I think we need some extra logic down there to see if `withReplacements` has a value.

iansvo: This is now updated

tomusborne (approved):

PR #1516, opened 2024-11-21
Fix query loop paging link issue

### LINKED ISSUE #1514: GB 2.0 Query loop paging links don't work for GP Block Element Loop Template
### Description
Pagination for Query Loop, with Inherit Query from Template in GPP  loop template  returns bad page link URL with query string 
eg. 
`https://my-site/archive/?query-712eb604-page=2 `

Instead of:

`https://my-site/archive/page/2`

Works correctly for Next and Previous links.

rsm0128:
closes #1514

### DIFF
--- a/src/blocks/query-page-numbers/block.json
+++ b/src/blocks/query-page-numbers/block.json
@@ -57,6 +57,7 @@
 		"generateblocks/queryType",
 		"generateblocks/query",
 		"generateblocks/queryId",
+		"generateblocks/inheritQuery",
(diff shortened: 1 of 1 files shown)


rsm0128:
@tomusborne I don't have access to the `@edge22/*` node modules, so couldn't test it on my local, but I'm sure this PR will solve the issue #1514

tomusborne:
Nice find. Thanks for contributing!

PR #1534, opened 2024-11-25
Improve post_per_page UX

### LINKED ISSUE #1527: Query Block - Per Page field needs better handling
### Description
The Per Page field can accept non numeric values with no error handling. 
If a user saves a negative value, non numberic value or no value then the arg will be saved and the query will break the page on the front end.

The field should be limited to a numeric value with an option to display all "items"

iansvo:
Fixes #1527
Related: https://github.com/tomusborne/generateblocks-pro/pull/769


tomusborne:
@iansvo I'm still able to save super negative numbers in the posts per page field, is that intended?

I also was never able to type in non-numeric values, even in the current release branch.

iansvo:
Are you able to enter in a non-number for a post query or a pro query?

The code here should force that input to be a number with a min of -1, so seems like maybe you're looking at pro. If so you need the related PR.

tomusborne:
@iansvo I'm not able to add a non-numeric value in the release branch, or this branch.

I'm still able to add a value like `-12345` in this branch.

iansvo:
Ok so apparently we need to handle the minimum in the onChange as well as in the attribute. I also saw that the Control was not quite working correctly with this when I did the onChange fix, which I assume is because the componentValue becomes unsynced with the value prop.

PR #1540, opened 2024-11-26
Bug/dynamic tags user author meta

### LINKED ISSUE #1538: Dynamic Tag - Author Meta add Display Name.
### Description
The dynamic tag accepts: `display_name` as a key but it's not listed in the Dynamic Tag meta options. Can we add it please.

iansvo:
Fixes #1538 

**Requires**
- https://github.com/EDGE22-Studios/components/pull/12
- https://github.com/tomusborne/generateblocks-pro/pull/772

### Fixed
- Author meta was not showing the "data" keys from wp_users
- User meta not populating at all due to invalid parameter type
- Tag supports was not properly detected when the tag changed
- Post record request was happening once before the tag was loaded
- Various minor bug fixes


tomusborne:
@iansvo Testing the user meta issue mentioned above. I'm still only seeing `generateblocks_onboarding` as the only available key when querying the current user. Should I be seeing more than that?

iansvo:
@tomusborne yes, part of the reason for that is because author "data" and user meta aren't really the same thing. The former is made up of the columns in wp_user vs the rows in wp_usermeta.

So for the author, i have the data keys (such as display_name) in the dropdown, but I wasn't sure if it made sense to also do that for users. 

Part of me thinks that having author/user meta tags is fuckin pointless. They are effectively the same thing, except the user in question can be "author of this post" vs a user you explicitly set. 

Let me know what you think here. At the very least we could add the wp_user columns to the user_meta also, which I think still works even though they aren't actually meta.

iansvo:
@tomusborne I just pushed an update to both PRs that adds these keys to the user section as well. I also needed to create a new user record endpoint so we can have more control like we do with posts. 

All of this is sorted, but there's now another components PR that goes with it: https://github.com/EDGE22-Studios/components/pull/12

tomusborne:
@iansvo I'm still not seeing any user meta populating the dropdown.

Using both PRs, and both are linked with the components PR.

tomusborne:
Here's a video of what I'm seeing. Looks like it's not firing on initial load - something else has to happen first.


https://github.com/user-attachments/assets/1d8bda01-5c4b-4106-b2ec-5322962ed636

iansvo:
@tomusborne I resolved the issue there. I had a flaw in the way I was handling the userRecordId. This now should work correctly. I'm going to continue additional testing for authors to make sure this is leveraging only that as well. Previously it was using the post record.

iansvo:
@tomusborne I've made sure that authors and users now use the same user-record endpoint. Previously authors were using the post record for their data, which seems kinda pointless now. 

This is all working in my tests. I can confirm both dropdowns populate as expected without any user interaction.

PR #1541, opened 2024-11-27
Refactor: Dynamic tag parser

### LINKED ISSUE #1539: Dynamic Tag- `{{post_date type:modified}}` does not render on frontend if `{{post_date}}` exists in same block.
### Description
If`{{post_date}}` and `{{post_date} type:modified}}` are added to a single block then the front end will only render the Post Date  

Screenshot 2024-11-26 at 14 21 06

tomusborne:
closes #1539 

Our current dynamic tag parser has a flaw that prevents multiple tags from being parsed if one of them has no options, and the other one does.

This PR aims to simplify the parser so that it treats all dynamic tags the same, regardless of whether they have options or not.

### DIFF
--- a/includes/dynamic-tags/class-register-dynamic-tag.php
+++ b/includes/dynamic-tags/class-register-dynamic-tag.php
@@ -47,7 +47,7 @@ class GenerateBlocks_Register_Dynamic_Tag {
 	 * @return array
 	 */
 	private static function parse_options( $options_string, $tag_name ) {
-		$pairs = explode( '|', $options_string );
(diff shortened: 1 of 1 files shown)


iansvo:
@tomusborne I pushed my commit with the changes that worked. Still haven't quite gotten the pipe handling down yet though.

tomusborne:
@iansvo Let's put those changes in a new PR based on this one so we can test further without it holding up this PR.

iansvo:
@tomusborne I reset the branch to the previous version and I opened a new PR #1543

PR #1543, opened 2024-11-28
Escaped character handling

iansvo:
Handles using quotes and pipe characters in dynamic tags.


tomusborne:
@iansvo I improved parsing here (fixed some bugs), and added JS parsing support.

We can now escape both `|` and `:` characters.

I also added some tests (PHP and JS).

PR #1547, opened 2024-11-29
Query - Update pagination and no results buttons

### LINKED ISSUE #1445: Query block - Pagination and No Results blocks can be added multiple times.
### Description
We should consider removing the Add Pagianation and Add No Results toolbar buttons if the blocks are already present inside the Query.

iansvo:
Fixes #1445

### DIFF
--- a/src/blocks/query/toolbar-appenders.js
+++ b/src/blocks/query/toolbar-appenders.js
@@ -1,9 +1,10 @@
 import { addFilter } from '@wordpress/hooks';
 import { createBlocksFromInnerBlocksTemplate } from '@wordpress/blocks';
 import { ToolbarButton } from '@wordpress/components';
-import { useDispatch } from '@wordpress/data';
+import { useDispatch, useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
 
+import { getInnerBlocks } from '@utils';
 import { PAGINATION_TEMPLATE, NO_RESULTS_TEMPLATE } from './templates';
 
 function noResultsIcon() {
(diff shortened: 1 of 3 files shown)


tomusborne:
@iansvo I've always thought that keep track of a blocks inner blocks causes a lot of re-renders/is an expensive check?

That's why I went this route with the caption: https://github.com/tomusborne/generateblocks/blob/release/2.0.0/src/blocks/media/toolbar-appenders.js#L54

Basically, if the user clicks on the toolbar button, we then do the checks to see if it exists. If it does, we select the block for the user.

iansvo:
@tomusborne Happy to change to whatever is most performant but looking at the code you've got here, this a.) doesn't seem like it'll work for me (I need inner blocks not children) and b.) You're basically doing the same thing as me at a different point (calling getBlock, which is what includes all the recursive innerBlocks data anyway). 

In this case, I'm only calling getBlock once, whereas in your code you're calling it each time on each block from the getBlocks call. As I type this and think about it, it kinda feels like my getInnerBlocks function would actually vastly simplify the code for your implementation and would be more performant because you're not calling getBlock on every single block parent and every single innerblock and just call it once instead.

```
if ( figure?.[ 0 ] ) {
  const innerBlocks = getInnerBlocks( figure[ 0 ] );
  const caption = innerBlocks.find( ( innerBlock ) =>
    'figcaption' === innerBlock.attributes.tagName
  );

  if ( caption ) {
    selectBlock( caption.clientId );
    return;
  }
}
```

getBlock will recursively include all innerBlocks, so I would think calling that less times would result in a performance boost of some sort.

iansvo:
Also, it seems like you might want to use [getBlockParentsByBlockName](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-block-editor/#getblockparentsbyblockname) so you don't have to do the conditional check if the block is an element or whatever.

tomusborne:
The main difference is that the code I linked to only runs when you click the button in the toolbar. We're not checking `innerBlocks` on render of the actual button itself, each time that button re-renders (which could potentially be a lot).

iansvo:
@tomusborne I pushed an update that uses the same innerBlocks comparison logic but I did it onClick as you've recommended. I've tested and the removal/block selection all works as expected. 

Let me know if you have any other feedback.

src/blocks/query/toolbar-appenders.js
```diff
@@ -28,8 +29,10 @@ function paginationIcon() {
 	);
 }
 
-function QueryToolbar( toolbars, { clientId, name } ) {
-	const { insertBlocks } = useDispatch( 'core/block-editor' );
+function QueryToolbar( toolbars, props ) {
+	const { clientId, name } = props;
+	const { insertBlocks, selectBlock } = useDispatch( 'core/block-editor' );
+	const getBlock = useSelect( ( select ) => select( 'core/block-editor' )?.getBlock );
```
tomusborne, line 35: Do we need this here?

tomusborne: Yea, I read it wrong. Looks good!

PR #1560, opened 2024-12-03
Query - fix issue with current handling

iansvo:


includes/class-query-utils.php
```diff
@@ -153,10 +121,19 @@ public function get_wp_query( $request ) {
...
 	 * @param WP_Block|array $block The current block.
+	 * @param array          $current Array of current entities (post, author, etc.).
 	 *
 	 * @return array $query_args The optimized WP_Query args array.
 	 */
-	public static function get_wp_query_args( $args = [], $page = 1, $attributes = [], $block = null ) {
+	public static function get_wp_query_args( $args, $page = 1, $attributes = [], $block = null, $current = [] ) {
+		// Bail if no args are provided.
+		if ( ! $args ) {
+			return [];
```
tomusborne, line 131: @iansvo Should we bail here? Prevents users from filtering the args of a query with no args.

PR #1564, opened 2024-12-04
Looper - Add block appender

### LINKED ISSUE #1444: Looper and Loop Item blocks can be deleted but not added back
### Description
The Looper and Loop item blocks should be locked by default to reduce the risk of accidental deletion.
We should also consider a toolbar option to add the blocks back if they have been deleted.

iansvo:
Fixes #1444


tomusborne:
@iansvo Is there a regular post query working for you when using this branch? I'm not seeing any posts come through in the editor.

PR #1569, opened 2024-12-05
Feature: Add `media` dynamic tag

tomusborne:
Requires https://github.com/EDGE22-Studios/components/pull/13

This allows us to use a `{{media}}` dynamic tag to pull in items from the Media Library.

This can be used to dynamic fetch images, or their captions/alt tags etc...

### DIFF
--- a/includes/dynamic-tags/class-dynamic-tags.php
+++ b/includes/dynamic-tags/class-dynamic-tags.php
@@ -793,6 +793,11 @@ class GenerateBlocks_Dynamic_Tags extends GenerateBlocks_Singleton {
 
 					if ( $media_id ) {
 						$processor->set_attribute( 'data-media-id', $media_id );
+
+						if ( ! $processor->get_attribute( 'alt' ) ) {
+							$processor->set_attribute( 'alt', get_post_meta( $media_id, '_wp_attachment_image_alt', true ) );
+						}
+
 						$content = $processor->get_updated_html();
 					}
 				}
--- a/includes/dynamic-tags/class-dynamic-tags.php
+++ b/includes/dynamic-tags/class-dynamic-tags.php
@@ -316,6 +316,30 @@ class GenerateBlocks_Dynamic_Tags extends GenerateBlocks_Singleton {
 				'return'   => [ 'GenerateBlocks_Dynamic_Tag_Callbacks', 'get_next_posts_page_url' ],
 			]
 		);
+
+		new GenerateBlocks_Register_Dynamic_Tag(
+			[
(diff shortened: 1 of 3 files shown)


iansvo (commented):
Overall looks pretty good, I haven't gotten to test it yet but I'll confirm after I do. 

Just one overall comment re: media selection.

iansvo (commented):
Based on the current change from select post, this may cause incorrect behavior in other places (ex: Query block options). 

We just need to make sure the other implementations are updated to use the correct postStatus value. The only 2 places that come to mind are the free/pro Query Inspector Controls.

includes/dynamic-tags/class-dynamic-tags.php
```diff
@@ -316,6 +316,30 @@ public function register() {
 				'return'   => [ 'GenerateBlocks_Dynamic_Tag_Callbacks', 'get_next_posts_page_url' ],
 			]
 		);
+
+		new GenerateBlocks_Register_Dynamic_Tag(
+			[
+				'title'    => __( 'Media', 'generateblocks' ),
+				'tag'      => 'media',
+				'type'     => 'media',
+				'supports' => [],
```
iansvo, line 325:
```suggestion
				'supports' => [ 'source' ],
```

src/dynamic-tags/components/DynamicTagSelect.jsx
```diff
@@ -722,6 +731,25 @@ export function DynamicTagSelect( { onInsert, tagName, selectedText, currentPost
 						</>
 					) }
 
+					{ 'media' === dynamicSource && (
```
iansvo, line 734: You need to update sourceOptions to use the correct labels for the media tag type.

includes/dynamic-tags/class-dynamic-tags.php
tomusborne: Do we need this though? Media only has the one source (the one you choose). It doesn't support things like `current`, or next/prev posts.

src/dynamic-tags/components/DynamicTagSelect.jsx
```diff
@@ -722,6 +731,25 @@ export function DynamicTagSelect( { onInsert, tagName, selectedText, currentPost
...
+						<>
+							<SelectPost
```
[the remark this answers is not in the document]
tomusborne: That would be ideal, but I wanted to get this into the release candidate, and I figured adding a new pattern like a Media Library modal would have some increased overhead.

It would definitely be nice for the user to be able to choose the image. Worth a quick play to see what kind of overhead we're looking at.

PR #1570, opened 2024-12-05
Dynamic Tags - Remove get_queried_object_id

iansvo:
Reference:  https://generate.support/topic/new-dynamic-tags-wont-render-on-front-end/#post-144381 

This was breaking the dynamic tags ID lookup when the dynamic tag was used in “content template” such as a loop item for an archive page.   Removing this has had no adverse effects, archive_title still works as expected, etc.


tomusborne:
Another report: https://generate.support/topic/featured_image-is-empty-for-gb-container-background/

tomusborne:
@iansvo What about things like the "Current" term for term meta?

tomusborne:
@iansvo I think this is the move: https://github.com/tomusborne/generateblocks/pull/1576

We actually already look for the `term` type when getting things like term meta: https://github.com/tomusborne/generateblocks-pro/blob/release/1.8.0/includes/extend/dynamic-tags/class-register.php#L245

PR #1571, opened 2024-12-06
Feature: Add alignment toolbar

### LINKED ISSUE #1463: Full Width and Wide Width Alignment options.
### Description
GB 2.0 does not, currently, have the core Full and Wide Width options.
This needs to be addressed as they are a must for Block Based themes.

Could we:

a) display them if Theme is a Block Theme
b) provide GenerateBlocks > Settings option to enable / disable them

tomusborne:
closes #1463 

alignment

This PR adds an alignment toolbar option.

For the Container block, it adds:

1. Text align
2. Block width (if supported by the theme)

For the Text block, it adds:

1. Text align

For the Media block, it adds:

1. Float


src/blocks/media/edit.js
```diff
@@ -171,6 +172,17 @@ function EditBlock( props ) {
 		<>
 			<InspectorControls>
 				<StylesOnboarder />
+
+				{ 'img' === tagName && (
```
iansvo, line 176: For future, I think this check may need to include picture elements once we get those dialed in. NCR for now.

tomusborne:
I've removed the image float stuff for now. There were some issues in the editor when floating, and we need to figure out what happens when a user wants to float an image that exists inside of a `<figure>`.

PR #1576, opened 2024-12-07
Fix: Taxonomy term dynamic tags

tomusborne:
Replaces https://github.com/tomusborne/generateblocks/pull/1570


iansvo:
Good catch here.

PR #1584, opened 2024-12-09
Implement new SelectTerms

iansvo:
Implement new SelectTerms component

Requires: https://github.com/EDGE22-Studios/components/pull/15
Related: https://github.com/tomusborne/generateblocks-pro/pull/786


tomusborne:
@iansvo Let's revert the changes here that remove `source` as a valid supports for the term meta.

iansvo:
I've restored the source functionality and confirm the "current term" option works as expected in elements.

tomusborne:
@iansvo Only issue I'm seeing here is that I'm unable to clear the Select Terms select using the `x` icon.

PR #1587, opened 2024-12-10
Tweak: Add responsive grid defaults

tomusborne:
This PR makes our Grid block variation responsive by default, and adds some default grid gap.


src/utils/mobileAtRule.jsx
iansvo, line 1: This file really shouldn't have a .jsx extension if it only returns data or a function. JSX extensions should be reserved for files that return JSX (like a component). 

There's a better pattern we can use for this kind of thing, I have it in my PR.

PR #1607, opened 2024-12-18
Feature - Add filter to disable dynamic tag previews in editor

### LINKED ISSUE #1583: Dynamic Tags - add option to disable previews in editor
### Description
Add option to disable previews which should solve for higher API request counts.

iansvo:
Closes #1583


src/dynamic-tags/components/DynamicTagBlockToolbar.jsx
```diff
@@ -51,9 +51,14 @@ export function DynamicTagBlockToolbar( {
 	onChange,
 	context,
 } ) {
+	const previewEnabled = 'enabled' === generateBlocksEditor?.dynamicTagsPreview;
 	const allTags = generateBlocksEditor.dynamicTags;
 	const foundTags = getTags( value, allTags );
 
+	if ( ! previewEnabled && 'preview' === contentMode ) {
+		setContentMode( 'edit' );
```
tomusborne, line 59: @iansvo I feel like these setters should be in an effect. One of the bugs we saw before (https://github.com/tomusborne/generateblocks/pull/1558) was caused by this kind of thing.

src/hoc/withDynamicTag.js
```diff
@@ -17,8 +17,13 @@ export function withDynamicTag( WrappedComponent ) {
 
 		const [ dynamicTagValue, setDynamicTagValue ] = useState( '' );
 		const [ contentMode, setContentMode ] = useState( 'edit' );
+		const previewEnabled = 'enabled' === generateBlocksEditor?.dynamicTagsPreview;
 		const isSavingPost = useSelect( ( select ) => select( 'core/editor' ).isSavingPost() );
 
+		if ( ! previewEnabled && 'preview' === contentMode ) {
+			setContentMode( 'edit' );
```
tomusborne, line 24: Same comment as the above.

tomusborne:
Aside from the comments above this looks good. Any reason you chose to go with `enabled` and `disabled` vs just letting the `true` or `false` of the filter come through?

iansvo:
@tomusborne When you add stuff in the inline script output like that it doesn't actually use true/false, it just sets the value to a number. So I figured this would be a bit more obvious than '1' or '0'.

src/dynamic-tags/components/DynamicTagBlockToolbar.jsx
iansvo: I've updated it, but I confess I'm a little unsure as to why that's a problem. This kind of thing is often done in the React docs even, I suspect it's because some of these are HOCs or some shit like that.

src/hoc/withDynamicTag.js
iansvo: Updated

PR #1608, opened 2024-12-18
Bug - Loop Item tag not always visible in Loop Item block

iansvo:
Closes https://github.com/tomusborne/generateblocks-pro/issues/785

### DIFF
--- a/src/blocks/looper/components/LoopInnerBlocksRenderer.jsx
+++ b/src/blocks/looper/components/LoopInnerBlocksRenderer.jsx
@@ -256,7 +256,7 @@ export function LoopInnerBlocksRenderer( props ) {
 		return ( <Spinner /> );
 	}
 
-	return hasResolvedData ? loopItemsContext.map( ( loopItemContext, index ) => {
+	return loopItemsContext.map( ( loopItemContext, index ) => {
 		// Include index in case the postId is the same for all loop items.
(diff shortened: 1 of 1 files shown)


tomusborne:
This seems to break the blocks inside of the Looper.


https://github.com/user-attachments/assets/82173f22-1910-4d76-a2c6-2df54d01ee3e

I'm also not 100% sure why the changes in the PR would fix the mentioned issue? Any context there?

iansvo:
This issue prevents the Loop Item tag from being hidden because the loop doesn't see anything. I had previously set up some fallback context, but the hasResolvedData check when rendering caused it to not get used. 

So by doing this, loopItemsContext _always_ wraps the output, no matter if the query is empty or not. This ensures there's always a loopItem, loopIndex, and previewId for loop items, even if the query is empty. 

I made a tweak just now to make sure the currentPostId was being used here, just in case. With this change, I'm seeing everything work as expected. Can you have another look and confirm?

tomusborne:
I'm still having issues selecting blocks with this PR.

To replicate:

1. Add Query
2. Start blank
3. Select Loop item and add empty text
4. Select something else and try to select text again

tomusborne:
I'm still seeing weirdness here.

https://github.com/user-attachments/assets/0b0b5ea5-d76c-4853-a178-f720c661ed85

At this point I'm thinking maybe we should just always show the Loop Item/Index options and ditch the conditional visibility until another date.

PR #1611, opened 2024-12-19
Add array support for dynamic tags outside of a query

### LINKED ISSUE #787: release 1.7 - Auto Margins on Headline and Image Blocks not working in editor. OK on  the frontend.
### Description
Auto Margins on Headline and Image Blocks not working in editor. 
OK on the frontend.

iansvo:
Closes #787


includes/dynamic-tags/class-dynamic-tag-callbacks.php
```diff
@@ -374,16 +416,13 @@ public static function get_post_meta( $options, $block, $instance ) {
 			return self::output( '', $options, $instance );
 		}
 
-		$value = GenerateBlocks_Meta_Handler::get_post_meta( $id, $key, true );
+		$single = empty( $options['list'] ) && empty( $options['join'] );
```
iansvo, line 419: @tomusborne This is really the main thing that need to be done for the other meta/option types to support this change. Should be a small diff (your favorite).

PR #1614, opened 2024-12-20
Fix: Image block error

tomusborne:


src/hoc/withHtmlAttributes.js
```diff
@@ -62,7 +62,7 @@ function shallowEqual( obj1, obj2 ) {
 
 const sanitizeAttributeValue = ( value ) => {
 	// Replace characters like &, <, >, " with their HTML entity equivalents
-	return value
+	return value.toString()
```
iansvo, line 65: We need to use something a bit more robust here. If you pass undefined or null to this it's going to error and an object/array isn't going to print properly. I am tempted to say this sanitization probably should only happen if the value is a string or a number (which is easily coerced into a string). Anything else is chancy and I'm not sure anyone can actually enter anything that isn't a string anyways.

iansvo: Marking this resolved because @tomusborne is sure with the type handling.

iansvo (approved):
If we're comfortable with the type handling, LGTM

iansvo (approved):
Updates look solid here, nice work

Issue #1616, opened 2024-12-20
Reduce network requests


iansvo:
This issue is an epic to track ways we can reduce network requests. 

- [x] https://github.com/tomusborne/generateblocks/pull/1619
- [x] https://github.com/EDGE22-Studios/components/pull/18
- [x] https://github.com/tomusborne/generateblocks-pro/pull/805

tomusborne:
Related: https://github.com/tomusborne/generateblocks/issues/1613

tomusborne:
Finding a way to reduce the requests for dynamic tag should be a priority. We can see a real-life example of the number of requests being a problem in the issue linked above.

I wonder if a simple local cache when editing the page would help.

iansvo:
I'm working on a POC for that right now. I'll report back when I have something to share.

iansvo:
@tomusborne So far I have a number of PRs that will positively impact this that affect Dynamic Tags, SelectPost, SelectUser, and Global Styles. 

The overall idea for each is the same: we implement tanstack query and leverage the caching mechanisms therein so we don't have to figure out if we need to make the request again or not. We have a QueryClientProvider that wraps key areas and passes the queryClient as a prop to those components so they can use the client. This allows the components to effectively remain just as portable, but now they can share the same caching logic. 

For instance, the Global Styles are the same when we request them, no matter which block it's for. So if we can cache that request, we don't have to ask for the global styles every single time we click on a new block. But if we create a new one, it will invalidate the query and we'll get a fresh result. Otherwise if a minute passes the request will be made again (staleTime). 

Let me know what you think on these PRs.

PR #1619, opened 2024-12-23
Dynamic Tags - Reduce and cache replacement requests

iansvo:
Relates to: https://github.com/tomusborne/generateblocks/issues/1616

This PR implements a local client side cache and server side object cache to reduce the amount of requests and database lookups.   Based on my tests it seems to be *much* faster afterwards. Please have a look and let me know what you see.


src/hoc/withDynamicTag.js
```diff
@@ -16,10 +41,15 @@ export function withDynamicTag( WrappedComponent ) {
 		} = attributes;
 
 		const [ dynamicTagValue, setDynamicTagValue ] = useState( '' );
-		const [ contentMode, setContentMode ] = useState( 'edit' );
+		const [ contentMode, setContentMode ] = useState( 'preview' );
```
tomusborne, line 44: This seems like it could have some side-effects that we don't want to introduce at this point. Was doing this necessary for the caching to work?

From a logic standpoint, I'm more comfortable with a block always starting in "edit" mode, and only switching to "preview" mode if it applies.

includes/dynamic-tags/class-dynamic-tags.php
```diff
@@ -499,15 +499,32 @@ public function register_rest_routes() {
 	 * @return WP_REST_Response
 	 */
 	public function get_dynamic_tag_replacements( $request ) {
-		$content      = urldecode( $request->get_param( 'content' ) );
```
tomusborne, line 502: Was this unnecessary? Maybe left over from when we were using a GET or something?

iansvo: @tomusborne Yes I believe this was leftover from the GET. Removing it didn't seem to change anything either.

src/dynamic-tags/components/DynamicTagBlockToolbar.jsx
```diff
@@ -74,12 +73,6 @@ export function DynamicTagBlockToolbar( {
 		return contentValue.substring( selectionStart.offset, selectionEnd.offset );
 	}, [ selectionStart, selectionEnd, value ] );
 
-	useEffect( () => {
-		if ( foundTags.length && ! isSelected ) {
-			setContentMode( 'preview' );
-		}
-	}, [ foundTags.length, isSelected ] );
```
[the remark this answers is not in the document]
iansvo: I restored this, I tried testing without it and I can't seem to find a reason we need it. I feel like I had a reason before, but again I'm not 100% sure.

src/hoc/withDynamicTag.js
```diff
@@ -43,14 +72,30 @@ export function withDynamicTag( WrappedComponent ) {
...
 			fetchData();
-		}, [ contentValue, contentMode, context, tagName, isSavingPost ] );
+		}, [
+			contentValue,
+			contentMode,
+			context,
+			tagName,
+			isSavingPost,
+			blockCacheKey,
+			isSelected,
```
tomusborne, line 97: Do we want/need this to run every time `isSelected` changes?

iansvo: I can test again to make 100% sure, but I feel like this was necessary to refresh the value in case it changed or something along those lines. I've done a few PRs since then so I don't remember offhand.

tomusborne (approved):

PR #1622, opened 2024-12-28
Query Block - Reduce extra network requests

iansvo:
Relates to: #1616 

In this branch, I add the QueryClientProvider to the plugin and pass the queryClient around so the consuming components can use it. This allows tanstack query to properly cache the queries and not have them happen so often. 

Requires components PR: https://github.com/EDGE22-Studios/components/pull/18


tomusborne:
Noticed that this broke the user select.


https://github.com/user-attachments/assets/04f1cce8-174a-493e-9d1c-8e4d33b50999

Fixed it easy enough, but should we worry about this if we forget to pass the query client? Could there be a backup perhaps?

iansvo:
I'm not sure there's a way we could do a fallback that doesn't involve writing the whole fetching logic twice. There's only so many places these components are even used, so it shouldn't be hard to manage which props they need. 

If anything, we could update the components to throw an error if queryClient is undefined, which they won't do now. this way we can obviously see we made a mistake while developing and fix it, etc.

Issue #1613, opened 2024-12-20
2.0 Performance


Kovas:
Since switching to GB 2.0, I’ve been regularly encountering 500 errors on my website. It’s a shared server, so other factors could have changed, or it might be a combination of issues. However, this problem only started after installing GB 2.0. My hosting provider suspects one of the plugins, but the plugin stack is identical to before, except for adding GB 2.0. It might be worth investigating this further.

tomusborne:
Can you share some more info here so we can investigate?

1. How long have you been seeing this errors? Throughout the entire alpha/beta, or since a specific version?
2. What are the specific errors? Your `error_log` file should log any `500` errors with their origin.

Kovas:
1. I would say this issue started right from the alpha.1 version.  

2. Unfortunately, I enabled DEBUG only about an hour ago, so here’s what I’ve gathered so far:  

   - **Database Error**
   I often get this error:
     ```
     Warning: mysqli_real_connect(): (HY000/1203): User *** already has more than 'max_user_connections' active connections in /public_html/wp-includes/class-wpdb.php on line 1982  
     User *** already has more than 'max_user_connections' active connections  
     ```

   - **Hosting Analytics**
     I observed 10.5k total requests in just the last hour while working on the project. Here's 500 Error Log:
     ![Error Log Image](https://github.com/user-attachments/assets/590900af-88f2-4451-a0f0-52ef595bc6a1)

   - **Frequent 500 Errors**
     Most of the errors seem to occur during these requests:
       - `POST /wp-json/generateblocks/v1/dynamic-tag-replacements?_locale=user HTTP/2` ~6200 in 6h
       - Some `POST /wp-json/generateblocks/v1/get-wp-query?_locale=user HTTP/2` ~85 in 6h

    - **Generate Cloud Error**
    ```
    Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>generatecloud</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the <code>init</code> action or later. Please see <a href="https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/">Debugging in WordPress</a> for more information. (This message was added in version 6.7.0.) in /public_html/wp-includes/functions.php on line 6114
    ```
Not sure if this helps, as this is probably the first time I’m checking logs/analytics. :)

iansvo:
[...]

PR #1630, opened 2025-01-02
Fix: Inline dynamic bgs

tomusborne:
Fixes an issue mentioned here: https://github.com/tomusborne/generateblocks/issues/1613#issuecomment-2567555165


iansvo (approved):
Nice, good cache here

PR #1639, opened 2025-01-10
Improve: Legacy style transforms

tomusborne:
This PR improves block transforms from v1 to v2. It will now transform legacy effects to the new styles object.


iansvo (commented):
Took a pass through the code, it seems in order, i'll need to do some more thorough testing in browser to confirm. 

Would it be feasible to get unit tests for this? It seems like the kind of thing that should be testable and valuable to test if we can.

PR #1655, opened 2025-01-14
Fix: Editor offset with limited posts

### LINKED ISSUE #1643: [GB 2.0 beta] Query loop with offset and one post per page renders invisible on the backend
### Description
Having a query with one post per page on the backend and an offset of one renders the query invisible on the backend. It renders in the front end but it's hard to visualize what happens in the backend since you would either need to remove the offset temporarily or set different number of posts per page, etc.

#### References: 
[Video](https://vimeo.com/1046038369/446cb1cdd4)

tomusborne:
closes #1643


iansvo (approved):
Solid fix, im surprised I didn't notice this sooner.

PR #1756, opened 2025-02-05
Add fallback option to featured image

tdrayson:
It's not easy without using a hook / filter to set the fallback image for the featured image.

This pull request adds a number field option to the featured image that will be used if the post doesn't have a featured image.


tomusborne:
Ah, it seems this PR was closed automatically when 2.0.0 was merged. Any chance you can re-open @tdrayson?

tdrayson:
It wouldn't let me re-open. So I've opened a new pull request - https://github.com/tomusborne/generateblocks/pull/1765

PR #1812, opened 2025-03-21
fix: update comments count option key to match callback expectation

alvindcaesar:
Fix comments count dynamic tag not using custom text for single comment.

The tag callback expects `single` as the option key, but registration used `one`. Updated registration to use correct key name.


tomusborne (approved):
Looks good to me! Nice find @alvindcaesar

